PR #51, opened 2026-05-12
fix(ui): restore editor focus after dismissing completion popup with Esc

narma:
## Summary

The SQL editor silently lost keyboard focus after pressing **Esc** to
dismiss the completion popup — focus_mode and the focus frame still said
"Editor", but typing did nothing.

This PR adds an `on_key_down` handler on the editor pane that re-focuses
the input on the next tick, and (as supporting infrastructure for
diagnosing this kind of bug on Windows) makes `env_logger` honour a
`DBFLUX_LOG_FILE` override.

## What does this resolve?

- A regression where pressing Esc to dismiss the SQL editor's completion
  popup would close the popup but break further input until the user
  clicked back into the editor.
- The closely related diagnostic gap: on Windows the GUI subsystem hides
  stderr, so reproducing focus bugs required other instrumentation.

## How was this solved?

**Root cause.** gpui-component's `CompletionMenu::hide` (invoked from
`InputState::escape` → `handle_action_for_context_menu`) only flips its
`open` flag and calls `cx.notify()`. The re-render that follows drops
`window.focus` before the next paint — even though the input still
owned focus synchronously when Esc was observed. Confirmed by logging
both phases:

```
[ESC-DEBUG] outer-pane key_down(escape): input_focused_now=true
[ESC-DEBUG] deferred refocus: input_focused_before_refocus=false
```

**Fix (`render_editor`).** Catch Esc in `on_key_down` and schedule a
re-focus via `cx.spawn_in`. Doing it inline does not help: `Window::focus`
early-returns when the handle is already focused, and by the time the
re-render clears it our synchronous call has already returned. The
deferred task runs on the next foreground tick and puts focus back.

**Logging change.** `env_logger::init()` now runs at the top of
`run_gui()` instead of after IPC/socket setup, and `DBFLUX_LOG_FILE`
redirects output to that file (append mode). Useful on Windows where
stderr is invisible under the GUI subsystem.

## Validation

- Reproduced on Windows: typing a letter to open the completion popup,
  then Esc to dismiss it — without the fix, the next character is
  dropped; with the fix, the editor stays usable.
- `DBFLUX_LOG_FILE` instrumentation logs (above) show the timing of the
  focus loss and confirm the deferred refocus restores it.
- `cargo build -p dbflux --features sqlite,postgres,mysql,mongodb,redis,dynamodb,aws`
- `cargo clippy -p dbflux -p dbflux_ui --lib -- -D warnings`
- `cargo test -p dbflux_ui --lib` — 398 / 398 pass

### Regression test note

I prototyped an integration test using a harness mirroring `render_editor`
(stub `CompletionProvider`, `simulate_input("S")`, `simulate_keystrokes("escape")`,
assert focus). It passes with the fix, but it also passes if the production
handler is reverted, because the harness owns its own copy of the
`on_key_down` body. Faithfully testing the production codepath would
require either standing up a full `CodeDocument` + `AppStateEntity` + a
fake connection so the menu actually opens, or extracting the handler
into a shared helper purely for tests. Both seemed disproportionate for
the size of the fix, so I dropped the test rather than ship a false
positive. Happy to revisit if a maintainer prefers one of those routes.

## Where was this tested?

- [x] Automated tests (the existing dbflux_ui suite)
- [ ] Linux
- [ ] macOS
- [x] Windows
- [ ] X11
- [ ] Wayland

## Checklist

- [ ] I added or updated tests when needed (see Regression test note above)

## Labels to apply

- `ui:bug`
- `platform:windows`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 All notable changes to DBFlux will be documented in this file.
 
+## [Unreleased]
+
+### Features
+
+* Logger now initialises at the very start of `run_gui()` so startup
+  diagnostics (IPC socket binding, auth token init) reach the log sink.
+  Setting `DBFLUX_LOG_FILE` redirects all `log::*!` output to the given
+  file in append mode — useful on Windows where the GUI subsystem hides
+  stderr.
+
 ## [0.5.0] – 2026-05-11
 
 ### Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,14 @@ All notable changes to DBFlux will be documented in this file.
   file in append mode — useful on Windows where the GUI subsystem hides
   stderr.
 
+### Fixes
+
+* SQL editor keeps focus after dismissing the completion popup with Esc.
+  gpui-component's `CompletionMenu::hide` clears the menu but the
+  follow-up re-render drops `window.focus` even though the input still
+  owned it synchronously; the editor pane now re-focuses its input on
+  the next tick so typing keeps working.
+
 ## [0.5.0] – 2026-05-11
 
 ### Features
--- a/crates/dbflux/src/main.rs
+++ b/crates/dbflux/src/main.rs
@@ -28,9 +28,39 @@ use interprocess::local_socket::{
     prelude::*,
 };
 use log::info;
+use std::fs::OpenOptions;
 use std::io::{self, Read, Write};
+use std::path::PathBuf;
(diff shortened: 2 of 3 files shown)


0xErwin1:
Thanks for the fixes!

narma:
@0xErwin1 I'm starting to use DBFlux at work, so I'm trying to fix the things that hinder or distract the most. Thank you for the wonderful project! Compared to DataGrip, it just flies :)

PR #62, opened 2026-05-13
fix(query): resolve drift-check schema via cache/editor, not hardcoded "public"

narma:
## Summary

Fixes a schema-drift preflight regression that fired a false "Schema has changed since this query was opened" modal — with **every column reported as `removed`** — for any unqualified query whose table lives outside the `public` schema. Clicking *Refresh & re-run* then poisoned the local cache with the empty `TableInfo`, blanking out autocomplete and the table-detail view until the user manually refreshed from the sidebar; the very next query reproduced the loop.

The fix is two-part:

1. Layered schema resolution in `check_schema_drift` so the fresh fetch is sent to the right schema.
2. A defensive guard that skips drift entries whose driver lookup returns zero columns, so the cache can never be poisoned through this path again.

Plus one repo-hygiene chore (`.gitattributes` pinning LF) that came up while preparing the patch on Windows.

## What does this resolve?

Reported in private review with a screenshot showing all 32 columns of a non-`public`-schema table marked `removed` against a PostgreSQL connection. Repro:

- The "Schema has changed" preflight modal misreports every column as removed when the query references a table outside `public` and is unqualified (e.g. `SELECT … FROM events` while the toolbar's Schema selector is on `analytics`).
- Subsequent "Refresh & re-run" overwrites the cached `TableInfo` with the empty result, breaking autocomplete and the column tree until a manual sidebar refresh.
- After a sidebar refresh, the next unqualified query reproduces the modal — endless cycle.

No upstream issue tracker reference.

## Root cause

`crates/dbflux_core/src/schema/drift_check.rs` (pre-fix line 85):

```rust
let schema = table_ref.schema.as_deref().unwrap_or("public");
```

The lightweight query parser at `crates/dbflux_core/src/schema/query_parser.rs` returns `QueryTableRef { schema: None, … }` for an unqualified `FROM events`. The drift checker then unconditionally falls back to `"public"`, calls `connection.table_details(db, Some("public"), "events")`, and the PostgreSQL driver's `get_columns` (`crates/dbflux_driver_postgres/src/driver.rs:2319`) issues:

```sql
SELECT … FROM pg_attribute a
JOIN pg_class c ON c.oid = a.attrelid
JOIN pg_namespace n ON n.oid = c.relnamespace
WHERE n.nspname = $1 AND c.relname = $2 …
```

For a table that lives in a non-`public` schema, this returns zero rows → `Ok(TableInfo { columns: Some(vec![]) })`. The fingerprint differs from the cached version (which holds the real columns from when the sidebar opened the table with the correct schema), so `diff_table_info` reports each column as `ColumnRemoved`. The modal's "Refresh & re-run" path at `crates/dbflux_ui/src/ui/document/code/execution.rs:1558-1562` then writes that empty `TableInfo` into `ConnectedProfile.table_details`, which is the same map autocomplete and `Sidebar::ensure_table_details` read from — hence the downstream blanking.

The drift checker also ignored the editor toolbar's `exec_ctx.schema` selector entirely, so even when the user *had* told the editor "I'm working in this schema", the drift preflight didn't know about it.

### `crates/dbflux_core/src/schema/drift_check.rs`

- Added a `default_schema: Option<&str>` parameter.
- Replaced the hardcoded `unwrap_or("public")` with a layered precedence — falling back to the cached entry's own `schema` field whenever a cached `TableInfo` exists, and to the caller-supplied `default_schema` otherwise:

  ```rust
  let schema = table_ref
      .schema
      .as_deref()
      .or_else(|| cached.and_then(|c| c.schema.as_deref()))
      .or(default_schema)
      .unwrap_or("public");
  ```

  The cached `TableInfo.schema` is the right authority because the only paths that populate the cache (`Sidebar::spawn_fetch_table_details`, `refresh_schema_object`) fetch with the schema from the sidebar tree node, where the schema is unambiguous.

- Added a defensive early-`continue` when the fresh fetch comes back with `columns: None` or `Some(vec![])`. A real table can't have zero columns; the much more likely interpretations are wrong-schema lookup, dropped table, or search-path race — none of which should produce a phantom diff or persist an empty `TableInfo` into the cache. Skipped entries get a `log::warn!` so the diagnostic is preserved.

### `crates/dbflux_ui/src/ui/document/code/execution.rs`

- `start_drift_preflight` now captures `self.exec_ctx.schema` (the toolbar's Schema dropdown) and forwards it as `default_schema` to `check_schema_drift`.

### `.gitattributes` (chore)

Pins `* text=auto eol=lf` plus a binary-extensions list. `rustfmt.toml` is configured `newline_style = "Unix"`, so `cargo fmt` writes LF; without `.gitattributes`, contributors on Windows whose global git config has `core.autocrlf=true` (the Git for Windows default) see hundreds of files flagged as modified after the first format run, even though `git diff --cached --stat` would be empty. Repo-level `eol=lf` removes the failure mode for everyone regardless of OS or local git config.

### What I actually ran

- `cargo check -p dbflux_core --tests` and `cargo check -p dbflux_ui --features "sqlite,postgres,mysql,mongodb,redis,dynamodb,aws"` — clean.
- `cargo clippy -p dbflux_core -- -D warnings` — clean.
- `cargo test -p dbflux_core --lib schema::` — 43/43 pass.

### Tests added (5 new integration tests in `drift_check.rs`)

A minimal `MockDriver` implementing `Connection` is added to the `mod tests` block. It captures every call to `table_details` so assertions can verify which schema was actually queried. Coverage:

| Test | What it locks in |
|---|---|
| `cached_schema_overrides_default_schema_for_fresh_fetch` | Reproduces the original bug — without the cached-schema fallback the test fails with `DriftOutcome::Drift` and the assertion message *"must not report drift when cached.schema steers fresh fetch to the right schema"*. |
| `default_schema_used_when_cache_is_empty_and_query_unqualified` | First-encounter lookup uses the caller's `default_schema` (the toolbar selection) instead of `"public"`. |
| `explicit_query_schema_wins_over_cached_and_default` | Sanity check that a qualified `FROM public.events` still routes to `public`. (Passes against the buggy code too — explicit qualifiers were never broken.) |
| `empty_fresh_columns_do_not_emit_diff_or_refresh` | Defensive guard: cached has columns, fresh returns empty → outcome is empty `Refresh`, not `Drift`. |
| `empty_fresh_columns_skipped_on_first_encounter` | Defensive guard: no cache entry, fresh returns empty → empty `Refresh`, never persists the empty `TableInfo`. |

I confirmed the suite is meaningful by reverting just the function body (keeping the new signature so it compiles) and re-running:

```
running 9 tests
test schema::drift_check::tests::cached_schema_overrides_default_schema_for_fresh_fetch ... FAILED
test schema::drift_check::tests::default_schema_used_when_cache_is_empty_and_query_unqualified ... FAILED
test schema::drift_check::tests::empty_fresh_columns_do_not_emit_diff_or_refresh ... FAILED
test schema::drift_check::tests::empty_fresh_columns_skipped_on_first_encounter ... FAILED
test schema::drift_check::tests::explicit_query_schema_wins_over_cached_and_default ... ok
…
test result: FAILED. 5 passed; 4 failed
```

With the fix restored: 9/9 pass.

### What I did NOT do

- I did not run the app interactively against a live PostgreSQL instance to confirm the bug disappears end-to-end. The bug report came from a user-provided screenshot of the modal, and the fix is locked in by tests that reproduce the exact failure path; I think that's sufficient evidence, but a maintainer with a quick repro setup may want to spot-check.
- I did not verify on Linux or macOS. The change is platform-independent in principle, and CI should cover the rest.

### Pre-existing failures noted but not addressed

- `cargo clippy --workspace -- -D warnings` fails inside `dbflux_proxy` with `clippy::large_enum_variant` on `enum ProxiedStream`. **Verified** on a clean checkout of `upstream/main` (stashed my changes, checked out upstream's `dbflux_proxy/`, ran clippy, got the same error). Not a regression from this branch.
- `cargo test -p dbflux_core --lib` reports 6 failures under `connection::hook::tests::execute_*` on Windows (subprocess / shell tests). I did **not** explicitly verify these on upstream's untouched `main` — I'm flagging them based on the test names being unrelated to my changes. If a maintainer wants confirmation either way, please ping and I'll bisect.

## Where was this tested?

- [x] Automated tests (5 new integration tests + the existing `check_drift_sync` suite)
- [ ] Linux
- [ ] macOS
- [x] Windows (build / test / clippy / fmt only — no manual UI verification)
- [ ] X11
- [ ] Wayland

## Checklist

- [x] I verified the change against the affected user flow(s) — via tests, not interactive UI
- [x] I documented follow-up work or known limitations when applicable (the cache key is `(database, table_name)` — schema-name collisions across same-named tables are still possible; out of scope for this fix and noted here for awareness)
- [x] I updated `CHANGELOG.md` under `## [Unreleased]` (one entry under `### Fixes`, one under `### Chores`)
- [x] I applied the appropriate labels (see below)

## Labels to apply

- `query:bug` (drift detection lives under the query subsystem)
- `driver:postgres` (most affected; the bug is reproducible against any driver where tables can live outside `public`)
- `kind:sql`
- `platform:windows` (only because the chore part — `.gitattributes` — fixes a Windows-specific contributor pain point; the bug fix itself is platform-independent)

---

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,33 @@
+# Normalize all text files to LF in the repo and in working trees.
+#
+# Without this, Windows contributors with `core.autocrlf=true` see CRLF in
+# their working tree while the repo (and tools like `cargo fmt`, which writes
+# LF per `rustfmt.toml`) speak LF. The first format run rewrites every file
+# from CRLF to LF in the working tree and `git status` flags every one as
+# modified — even though committing would be a no-op. Pinning `eol=lf`
+# repo-wide removes that mismatch for everyone.
+
+* text=auto eol=lf
+
+# Common binary types — never touch.
+*.png   binary
+*.jpg   binary
+*.jpeg  binary
+*.gif   binary
+*.ico   binary
+*.icns  binary
+*.webp  binary
+*.svg   text eol=lf
+*.woff  binary
+*.woff2 binary
+*.ttf   binary
+*.otf   binary
+*.zip   binary
+*.gz    binary
+*.tar   binary
+*.7z    binary
+*.pdf   binary
+*.dylib binary
+*.so    binary
+*.dll   binary
+*.exe   binary
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,23 @@ All notable changes to DBFlux will be documented in this file.
 
 ## [Unreleased]
 
+### Fixes
+
+* Schema-drift preflight no longer reports phantom "all columns removed"
+  for queries whose table lives outside `public`. The fresh fetch is now
+  steered to the right schema via a layered precedence (query qualifier
+  → cached `TableInfo.schema` → editor toolbar's schema → `public`
+  fallback), and the checker defensively skips any entry whose driver
+  lookup returns zero columns — preventing the empty `TableInfo` from
+  poisoning the autocomplete and table-detail caches via the
+  "Refresh & re-run" path.
+
+### Chores
+
+* Pin EOL to LF via `.gitattributes` (`* text=auto eol=lf`) so
+  `cargo fmt` no longer desyncs Windows working trees that default to
+  `core.autocrlf=true`.
+
 ## [0.6.0-dev.0] - 2026-05-12
 
 ### Features
--- a/crates/dbflux_core/src/schema/drift_check.rs
+++ b/crates/dbflux_core/src/schema/drift_check.rs
@@ -50,9 +50,30 @@ pub fn check_drift_sync(
 /// Check whether any tables referenced by `query` have drifted since they were
 /// last cached in `table_details`.
 ///
+/// `default_schema` is the caller's hint for which schema to introspect when
+/// the query does not qualify the table (e.g. the editor's active schema
+/// selector). It is only consulted as a tertiary fallback — the cached
+/// `TableInfo.schema` takes precedence because it records the exact schema
+/// the table was originally loaded from.
+///
+/// **Schema resolution precedence per table reference:**
+/// 1. Explicit qualifier in the query (`FROM public.users`).
+/// 2. The cached entry's `schema` field, when an entry exists. The cache is
+///    populated by sidebar/refresh paths that always know the right schema,
+///    so the cached entry is the authoritative pointer to where the table
+///    actually lives.
+/// 3. The caller-supplied `default_schema` (toolbar selection).
+/// 4. `"public"` as a last-resort driver-friendly fallback.
+///
 /// **Trigger rules:**
 /// - If the driver's `referenced_tables` returns `None`, return `Skip`.
 /// - For each referenced table:
+///   - If the fresh fetch returns no columns at all, skip the entry silently.
+///     A real table with zero columns is essentially impossible; an empty
+///     result almost always means the schema lookup was wrong (search path,
+///     dropped table, search-path race). Reporting "all columns removed"
+///     would be noise, and writing the empty `TableInfo` into the cache
+///     would poison autocomplete and the table-detail view.
 ///   - If no cache entry exists, this is a first encounter — add to `refreshes`
 ///     and proceed silently (no diff possible).
 ///   - If the entry exists and fingerprints match, add to `refreshes` for a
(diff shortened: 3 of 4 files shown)


0xErwin1:
Thanks for the thorough write-up and the reproduction tests, this is a really clean fix. The layered precedence (query qualifier → cached `TableInfo.schema` → toolbar default → `public`) is the right call, and verifying the suite by reverting just the body to confirm 4/5 new tests fail is the kind of validation I wish more PRs had.

One piece of feedback for next time: the `.gitattributes` chore would have been better as its own PR. It is unrelated to the drift-check fix, and keeping unrelated changes split makes the commit history easier to bisect and revert if either piece misbehaves later. Not blocking here, just something to keep in mind going forward.

PR #64, opened 2026-05-14
feat(ui): platform-aware keybindings (Cmd on macOS, Ctrl elsewhere)

### LINKED ISSUE #63: Question: should macOS use Cmd as the primary modifier instead of Ctrl?
## Summary

On macOS the platform convention for application shortcuts is **Cmd**
(Cmd+S, Cmd+W, Cmd+N, Cmd+Enter, …), while DBFlux currently uses **Ctrl**
uniformly across all platforms. Should we adopt the macOS convention for
application-level shortcuts and keep Ctrl on Linux/Windows?

This question was first raised in #59 (which introduced the only
platform-aware binding so far — Cmd+C / Ctrl+C for `ResultsCopyCell`) and
deferred for a wider discussion. Opening this issue to settle the
direction before more bindings drift one way or the other.

## Problem

The current keymap (`crates/dbflux_ui/src/keymap/defaults.rs`) uses Ctrl
as the primary modifier on every platform. On macOS that means, for
example:

- `Ctrl+Enter` — run query
- `Ctrl+S` / `Ctrl+Shift+S` — save / save as
- `Ctrl+N`, `Ctrl+W`, `Ctrl+1..9` — tab management
- `Ctrl+P` — saved queries
- `Ctrl+O` — open script file
- `Ctrl+B` — toggle sidebar
- `Ctrl+E` — export results
- `Ctrl+Shift+P` — command palette
- `Ctrl+Shift+A` — audit viewer

This is unusual for a native macOS app. Users coming from other database
clients, editors (VS Code, Zed, JetBrains), and the OS itself expect Cmd
here.

The codebase is already inconsistent on this point: `ResultsCopyCell`
uses Cmd+C on macOS via a `cfg(target_os = "macos")` branch (#59), and
the embedded data-table component uses `secondary-c` for Copy — so a few
shortcuts already follow platform convention while the rest do not.

## Alternatives

Posting these as a set, not a recommendation:

### A. Adopt platform-correct modifiers everywhere

Treat #59 as the start, and migrate the rest. A "primary" modifier
abstraction (e.g. `Modifiers::primary()` resolving to Cmd on macOS / Ctrl
elsewhere, plus the existing GPUI `secondary-*` form at binding sites)
keeps the bulk of the keymap declarative.

Carve-outs that would likely stay literal-Ctrl on every platform:

- Vim-style panel navigation (`Ctrl+h/j/k/l`, `Ctrl+u/d`) — Ctrl is the
  vim convention regardless of OS.
- `Ctrl+Tab` / `Ctrl+Shift+Tab` — universal tabbed-UI idiom; `Cmd+Tab`
  is the macOS system app switcher.
- Combinations that clash with macOS system shortcuts — `Cmd+M`
  (minimize), `Cmd+Shift+3` / `Cmd+Shift+4` (screenshots), etc.

Pro: native macOS feel, matches user expectation.
Con: muscle memory no longer fully portable between Mac and
Linux/Windows; carve-outs add nuance reviewers need to remember.

### B. Stay on Ctrl everywhere

Revert the partial Cmd handling (including the cfg gate from #59) and
document DBFlux as intentionally Ctrl-first across all platforms — like
some Linux-leaning tooling does.

Pro: uniform codebase, portable muscle memory, no carve-outs.
Con: macOS users keep pressing Ctrl in an app where every other native
tool uses Cmd; mixed-fleet teams will likely report this as a bug.

### C. Make the modifier configurable

Ship a platform-aware default (option A) but expose a setting — e.g. in
`keymap.json` or the app config — that lets the user pick `"platform"`
vs `"ctrl"` for the primary modifier. Mac users who want portable muscle
memory can opt out; new Mac users get the convention without thinking.

Pro: serves both camps; doesn't lock in a choice.
Con: more surface area (config schema, docs, command-palette labels need
to reflect the resolved modifier, tests cover both modes); risk that
"default" becomes an effective decision anyway.

## Open Questions

- If we go with **A** or **C**, where should the seam live — a helper on
  `Modifiers` in `dbflux_app::keymap::chord`, GPUI's existing
  `secondary-*` prefix at binding sites, or per-call
  `#[cfg(target_os = "macos")]`?
- How do we keep user-visible labels in sync — command-palette hints,
  tooltips, docs — with the platform-resolved chord?
- Scope: ship as one batch, or stage by command group (run query →
  save → tabs → navigation)?

## Validation / Impact

- Affects every macOS user; no behaviour change on Linux/Windows for
  platform-resolved bindings.
- Verifiable end-to-end on macOS (run query, save, palette, tabs, copy)
  plus a Linux/Windows smoke pass to confirm no regression.
- Existing tests in `crates/dbflux_ui/src/keymap/defaults.rs` would need
  platform-aware coverage.

---

I use DBFlux on both Windows and macOS, and personally find Cmd on macOS
/ Ctrl on Windows the most natural setup — so I lean toward **A** (or
**C** with A as the default).

Happy to open a PR for whichever direction we land on (A, B, or C).

narma:
## Summary

Application-level keyboard shortcuts now use Cmd on macOS and Ctrl on
Linux/Windows, following platform conventions. The seam is a new
`Modifiers::primary()` helper that collapses the per-call cfg pattern
introduced in #59 for the Results copy-cell binding.

vim-style navigation, `Ctrl+Tab`, and a handful of shortcuts that would
clash with system bindings on macOS (Cmd+M minimize, Cmd+Shift+3/4
screenshots) stay literal Ctrl on every platform.

## What does this resolve?

- Resolves #63

## How was this solved?

Five focused commits, each one a logical step:

1. **`feat(keymap): platform-aware primary modifier for app commands`**
   Adds `Modifiers::primary()` / `primary_shift()` in
   `dbflux_app::keymap::chord` (Cmd on macOS, Ctrl on Linux/Windows) and
   routes the app-level bindings through them in
   `dbflux_ui::keymap::defaults` + the Input-context override in
   `keymap::actions`. The existing `#[cfg(target_os = "macos")]` branch
   from #59 collapses into the helper. `KeyChord`'s `Display` order
   swaps to Cmd-first so labels read `Cmd+Shift+P` rather than
   `Shift+Cmd+P`.

2. **`feat(data-table): use secondary modifier for system-standard chords`**
   In `data_table::init`, the system-standard bindings (SelectAll,
   SaveRow, Undo, Redo) switch to GPUI's `secondary-*` modifier (Cmd on
   macOS, Ctrl elsewhere). Copy already used this. Vim aliases (`u`,
   `ctrl-r` Redo) and local mnemonics (`ctrl-n` SetNull) stay literal.

3. **`fix(ui): platform-aware editor and save-row shortcut hints`**
   The SQL editor toolbar caption and the data-grid save-row caption
   are cfg-gated so they read `Cmd+Enter` / `Cmd+↵` on macOS and
   `Ctrl+Enter` / `Ctrl+↵` elsewhere.

4. **`fix(ui): platform-aware palette shortcut labels`**
   `Workspace::default_commands()` builds the palette descriptor list
   with cfg-gated shortcut labels so macOS shows Cmd-prefixed strings.

5. **`fix(ui): kebab-case palette shortcut labels for chord rendering`**
   Per maintainer feedback: `palette_shortcut_parts` splits on `-` to
   build a multi-badge `Chord`. The previous `+`-form labels (`"Ctrl+N"`)
   were not split and rendered as one collapsed token. This commit
   switches the labels to kebab-case (`cmd-n` / `ctrl-n`, `ctrl-tab`,
   `ctrl-shift-1`, `esc`) so chords render as `[Ctrl] + [N]`.

Carve-outs that stay literal Ctrl on every platform — matching the
maintainer's note in #63 plus a few extras grounded in macOS system
conventions:

| Carve-out | Why |
|---|---|
| `Ctrl+h/j/k/l` focus nav, `Ctrl+u/d` page nav | vim/tmux idiom, as flagged on the issue |
| `Ctrl+Tab` / `Ctrl+Shift+Tab` | tabbed-UI idiom, as flagged on the issue (Cmd+Tab is the macOS app switcher) |
| `Ctrl+Shift+1..4` focus shortcuts | Cmd+Shift+3/4 are reserved by macOS for screenshots |
| `Ctrl+M` tab menu | Cmd+M is the system "minimize window" shortcut on macOS |
| History-modal local mnemonics (`Ctrl+D`/`Ctrl+F`/`Ctrl+R`) | Reusing Cmd here would clash with system Find/Reload |
| Data-table `ctrl-n` SetNull, `ctrl-r` Redo, `u` Undo | local mnemonics / vim aliases, not system-standard chords |

## Validation

```
cargo fmt --all -- --check        # ok
cargo check --workspace            # ok
cargo test -p dbflux_app keymap    # 12 passed
cargo test -p dbflux_ui --lib keymap        # 12 passed
cargo test -p dbflux_ui --lib command_palette  # 8 passed
```

New tests added in this branch:

- `chord::tests::test_display_cmd_first` — Display puts Cmd before Shift
- `chord::tests::test_primary_modifier_per_platform` — `primary()` resolves
  to `platform` on macOS, `ctrl` elsewhere
- `keymap::defaults::tests::macos_results_copy_uses_cmd_not_ctrl` —
  literal Ctrl+C does not trigger Results copy-cell on macOS

The pre-existing
`input_context_keybindings_override_secondary_enter` test was
generalized to run on macOS too (it had been gated to `not(macos)`).

## Where was this tested?

- [x] Automated tests
- [ ] Linux
- [x] macOS (manual smoke testing by author)
- [x] Windows (manual smoke testing by author)

## Known limitation (pre-existing, tracked separately)

While smoke-testing on macOS and Windows it was confirmed that
`Ctrl+Shift+1..4` (the direct focus shortcuts) don't move focus. This
pre-dates this PR — the bindings have always been the same on every
platform and the chord never matches the runtime keystroke. Root cause:
GPUI normalizes `Shift+digit` chords away from the digit on macOS and
Windows (intentionally, for non-US layouts), so the keystroke that
reaches `Workspace::on_key_down` doesn't match the
`KeyChord("N", Modifiers::ctrl_shift())` stored in `KeymapStack`.

Filed as a separate issue with the full root-cause analysis and three
alternative fixes: **#65**. Keeping it out of this PR so the
platform-modifier work stays reviewable on its own and the fix can get
its own platform-specific test coverage.

## Checklist

- [x] I applied the appropriate labels (see below)

## Labels to apply

- `ui:feature` — application-level keyboard shortcut conventions
- `platform:macos` — the change is observable on macOS specifically
- `platform:windows` — labels stay correct on Windows too

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,22 @@ All notable changes to DBFlux will be documented in this file.
 
 ## [Unreleased]
 
+### Changed
+
+* **Platform-aware keybindings** — application-level shortcuts now use
+  Cmd on macOS and Ctrl on Linux/Windows: command palette
+  (`Cmd/Ctrl+Shift+P`), new/close/switch tab, run query, save, open
+  script/history, export results, toggle sidebar, audit viewer, and
+  Results cell copy. vim-style navigation (`Ctrl+h/j/k/l`, `Ctrl+u/d`)
+  and `Ctrl+Tab` / `Ctrl+Shift+Tab` stay literal Ctrl on every platform,
+  along with focus shortcuts (`Ctrl+Shift+1..4`) that would clash with
+  macOS screenshot bindings and `Ctrl+M` which would clash with
+  window-minimize on macOS. Inline data-table commands (Copy, Save row,
+  Select all, Undo/Redo) use GPUI's `secondary-` modifier so they pick
+  the right key per platform. Command-palette shortcut labels and the
+  SQL editor / save-row hints now reflect the platform modifier. Closes
+  #63.
+
 ## [0.6.0-dev.0] - 2026-05-12
 
 ### Features
--- a/crates/dbflux_app/src/keymap/chord.rs
+++ b/crates/dbflux_app/src/keymap/chord.rs
@@ -43,6 +43,43 @@ impl Modifiers {
         }
     }
 
+    /// Platform-aware "primary" modifier: Cmd on macOS, Ctrl elsewhere.
+    ///
+    /// Use this for application-level commands (palette, save, copy, new tab,
+    /// run query, …) where the convention is Cmd on macOS but Ctrl everywhere
+    /// else. For vim-style navigation (Ctrl+hjkl, Ctrl+u/d) and bindings that
+    /// would clash with macOS system shortcuts (Cmd+M, Cmd+Shift+3/4), keep
+    /// using [`Modifiers::ctrl`] / [`Modifiers::ctrl_shift`] instead.
+    pub fn primary() -> Self {
+        #[cfg(target_os = "macos")]
+        {
+            Self {
+                platform: true,
+                ..Default::default()
+            }
+        }
+        #[cfg(not(target_os = "macos"))]
+        {
+            Self::ctrl()
+        }
+    }
+
+    /// Primary + Shift: Cmd+Shift on macOS, Ctrl+Shift elsewhere.
+    pub fn primary_shift() -> Self {
+        #[cfg(target_os = "macos")]
+        {
+            Self {
+                platform: true,
+                shift: true,
+                ..Default::default()
+            }
+        }
+        #[cfg(not(target_os = "macos"))]
+        {
+            Self::ctrl_shift()
+        }
+    }
+
     #[allow(dead_code)]
     pub fn has_any(&self) -> bool {
         self.ctrl || self.alt || self.shift || self.platform
@@ -128,8 +165,13 @@ impl KeyChord {
 
 impl fmt::Display for KeyChord {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // Order: Cmd, Ctrl, Alt, Shift — matches the convention used in
+        // user-facing docs (e.g. "Cmd+Shift+P", "Ctrl+Shift+P").
         let mut parts = Vec::new();
 
+        if self.modifiers.platform {
+            parts.push("Cmd");
+        }
         if self.modifiers.ctrl {
             parts.push("Ctrl");
         }
@@ -138,9 +180,6 @@ impl fmt::Display for KeyChord {
         }
         if self.modifiers.shift {
             parts.push("Shift");
-        }
-        if self.modifiers.platform {
-            parts.push("Cmd");
         }
 
         let key_display = match self.key.as_str() {
@@ -215,5 +254,33 @@ mod tests {
     fn test_display() {
         let chord = KeyChord::new("p", Modifiers::ctrl_shift());
         assert_eq!(chord.to_string(), "Ctrl+Shift+p");
+    }
+
+    #[test]
+    fn test_display_cmd_first() {
+        let chord = KeyChord {
+            key: "p".to_string(),
+            modifiers: Modifiers {
+                platform: true,
+                shift: true,
+                ..Modifiers::none()
+            },
+        };
+        assert_eq!(chord.to_string(), "Cmd+Shift+p");
+    }
+
+    #[test]
+    fn test_primary_modifier_per_platform() {
+        let m = Modifiers::primary();
+        #[cfg(target_os = "macos")]
+        {
+            assert!(m.platform);
+            assert!(!m.ctrl);
+        }
+        #[cfg(not(target_os = "macos"))]
+        {
+            assert!(m.ctrl);
+            assert!(!m.platform);
+        }
     }
 }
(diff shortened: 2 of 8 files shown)


0xErwin1:
Reviewed the full diff. This is a faithful implementation of option A and covers everything from #63: `Modifiers::primary()` as the single seam, the #59 cfg branch collapsed into it, every app-level binding migrated, and the carve-outs (vim nav, `Ctrl+Tab`, `Ctrl+Shift+1..4`, `Ctrl+M`) preserved with clear rationale.

Two things I especially appreciated. The smoke pass caught a pre-existing bug (`Ctrl+Shift+1..4` not firing) and you filed it as #65 instead of pulling it into scope. And commit 5 quietly fixes the palette badges not splitting into chords, which was a real display bug before this.

Thanks for the thorough work and the clear commit-by-commit breakdown. Merging with squash.

PR #172, opened 2026-06-05
fix(query-builder): materialize bound literals in SQL previews

0xErwin1:
## What

Visual builder previews and the mutation confirmation dialog showed bound literals as `?` instead of their values, e.g. `SET \`auto_retr_count\` = ?`. This made it look like the value wasn't applied, and for an UPDATE with a WHERE the preview mixed inlined and placeholder values.

## Why

The three preview sites rendered the raw parameterized SQL (`m.sql` / `q.sql`). Execution was always correct (params are bound at run time via `MutationExecutor`), but the preview misrepresented what would run.

## Change

- Add `GeneratedMutation::materialize_for_editor`, mirroring the existing `SelectQuery` method, backed by a shared `materialize_placeholders` helper (no logic duplicated).
- Use it in all three display-only preview sites: builder SELECT preview, builder mutation preview, and the confirmation dialog.
- Materialization is display-only; execution still uses `sql` + `params`, so bound parameters and the dangerous-query gate are unaffected.

## Notes

- Literals typed in the builder arrive as `ScalarLiteral::Text`, so a numeric value previews quoted (`= '1'`). Execution relies on the driver's implicit cast. Making numeric inputs derive a numeric `ScalarLiteral` is a separate change.

## Tests

- New `mutation_materialize_inlines_set_and_where_literals` in `dbflux_core` locks in `SET x = '1' WHERE id = 7`.
- Existing generator + materialization tests pass (`cargo test -p dbflux_core --lib query::generator`).

### DIFF
--- a/crates/dbflux_core/src/query/generator.rs
+++ b/crates/dbflux_core/src/query/generator.rs
@@ -66,41 +66,57 @@ impl SelectQuery {
     /// - `$N` placeholders (PostgreSQL): each `$N` is replaced by `params[N-1]`.
     /// - `@pN` placeholders (SQL Server): each `@pN` is replaced by `params[N-1]`.
     pub fn materialize_for_editor(&self, dialect: &dyn crate::sql::dialect::SqlDialect) -> String {
-        use crate::sql::dialect::PlaceholderStyle;
+        materialize_placeholders(&self.sql, &self.params, dialect)
+    }
+}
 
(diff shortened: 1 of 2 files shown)


0xErwin1:
Superseded by #176, which now targets main and bundles this preview fix together with the execution and typing fixes for the same builder bug. Closing to keep a single review.

PR #298, opened 2026-07-12
fix(hooks): preserve hooks and run test phases

### LINKED ISSUE #295: [bug] Hooks are not persisted, all load as empty Command after restart
### DBFlux version

latest git (74190d89)

### Operating system

Linux (Wayland)

### OS / distro detail

Arch Linux

### Architecture

amd64 / x86_64

### Affected driver (if any)

_No response_

### Steps to reproduce

Bug 1 (UI):
1. Settings > Hooks, create a new hook.
2. Set the type dropdown to Lua or Script.
3. The form does not change. No Lua or Script fields appear (file path, capabilities).
4. Save. The hook is stored as Command.

<details><summary>Possible fix</summary>
<p>

Syncs the value in the subscription in [hooks_section.rs:228-234](https://github.com/0xErwin1/dbflux/blob/74190d89b72d0269f783bff7ba83ba6cc6ff8df4/crates/dbflux_ui_windows/src/settings/hooks_section.rs#L228-L234)

``` rust
 let hook_kind_sub = cx.subscribe_in(
      &hook_kind_dropdown,
      window,
      |this, dropdown, _: &DropdownSelectionChanged, window, cx| {
          this.hook_kind_selection = match dropdown
              .read(cx)
              .selected_value()
              .map(|value| value.to_string())
              .as_deref()
          {
              Some("script") => HookKindSelection::Script,
              #[cfg(feature = "lua")]
              Some("lua") => HookKindSelection::Lua,
              _ => HookKindSelection::Command,
          };
          this.refresh_hook_script_content_editor(window, cx);
          cx.notify();
      },
  );
```

</p>
</details> 

Bug 2 (Persistence):
1. Create any hook (Command, Script, or Lua) and bind it to a profile phase.
2. Trigger the phase. The hook runs.
3. Restart dbflux.
4. Trigger the phase again. Nothing runs. The hook loaded as an empty Command.

### Expected behavior

1. Switching the type dropdown to Lua or Script shows the matching fields and is saved.
2. A global hook survives a save and reload, including its kind, script or file path, and Lua capabilities,
and runs at the phase it is bound to.

### Actual behavior

1. The type dropdown is not written into the form state
2. The hooks are not persisted correctly, shown as "empty" Command after restart

### Logs / stack traces

```shell

```

### Labels I plan to apply

- [x] Kind label (e.g. `driver:bug`, `ui:bug`, `query:bug`, `mcp:bug`, `rpc:bug`, …)

0xErwin1:
## Summary

- Synchronizes hook dropdown selections with editor state so changing Command, Script, or Lua updates the value that is actually saved.
- Preserves global definitions and profile bindings losslessly across restart, including Command, Script, and Lua configuration, with canonical migrations and protection for unreadable legacy rows.
- Runs PreConnect and PostConnect hooks during Test Connection and performs scoped cleanup without mutating the normal connection state.

## What does this resolve?

- Resolves #295

## How was this solved?

The change is organized into reviewable work units:

1. **Storage and migration commits:** introduce canonical hook-kind persistence, atomically replace definitions, retain durable identities, preserve profile variants, and protect legacy or unreadable rows from destructive rewrites.
2. **UI synchronization commits:** keep the Settings and Connection Manager dropdown/editor models aligned so persisted values match the visible selection.
3. **Hook runner commits:** extract the shared phase runner into `dbflux_ui_base`, then use it for Test Connection PreConnect and PostConnect execution with terminal-path and cleanup coverage.
4. **Verification commits:** add persistence, runtime-boundary, orchestration, cleanup, selection-parity, diagnostics, and workspace lint coverage.

This PR intentionally does not run disconnect phases from Test Connection, mutate normal connection state, or redesign the generic connection lifecycle, Tokio runtime ownership, or MCP integration.

The diff exceeds the normal review budget. The maintainer-approved `size:exception` applies; reviewers can follow the work-unit/commit sequence above rather than reviewing the full diff as one undifferentiated change.

## Validation

- PASS: `cargo fmt --all -- --check`
- PASS: `cargo check --workspace`
- PASS: `cargo clippy --workspace -- -D warnings`
- PASS: `cargo nextest run --workspace`, 4,228 tests passed via two complementary count partitions
- PASS: `cargo test --doc --workspace`, 3 passed and 10 ignored
- PASS: final SDD verification, 18/18 with no critical findings

## Where was this tested?

- [x] Automated tests
- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] X11
- [ ] Wayland
- [ ] Other:

No manual GUI smoke testing was performed.

## Checklist

- [x] I did not manually edit `CHANGELOG.md`; user-visible conventional commit messages feed git-cliff per `CONTRIBUTING.md`

## Labels to apply

- `ui:bug`
- `storage:bug`
- `size:exception`
- `rust`
- `status:needs-review`


x1h0:
Just tested your changes, it seems it does not cleanup broken hook definitions in the database where kind_json is null. Those old rows don't show up in the UI anymore, but they still keep the name and id, so I couldn't delete them or reuse the name. Had two of them and had to remove them directly in SQLite.
Maybe the fix could clean those up too? Or at least make them show up in the UI so they can be edited or deleted there.

Also hit another one with the connection hooks: when I pick hooks on a connection they don't get saved either, and they're gone after a restart. Saving the profile fails with:
`ERROR dbflux_app::app_state: Failed to save connection profiles: storage sqlite error for dbflux.db: FOREIGN KEY constraint failed`

x1h0:
I tested it again. It seems the hooks are working via "Test connection".
Unfortunately, I still can't set a hook for a connection; when I save, no log is generated, and the following message appears in the UI:
dbflux-error

0xErwin1:
Thanks for retesting, that error pinned it down. The dropdown now stores the hook's durable id instead of its name, but the save-time validation was still checking that id against the definition map keyed by name, so a valid selection always came back as `Unknown pre-connect hook ID` and blocked the profile save. Fixed in `db13a0bd`: it now validates the selected/typed tokens against the known id+name set, so real selections pass and only genuinely unknown entries are flagged.

Heads up: I haven't had time to run this through the UI myself yet, so it may still misbehave in a case I haven't hit. If you get a chance to pull the branch and try binding a hook again, let me know how it goes.

x1h0:
Okay now Hook & Connection saving works fine, but it seems hooks are not executed:
`2026-07-15T14:24:05.733551Z  WARN dbflux_core::connection::hook: Profile 'Test' references missing pre-connect hook 'f8416e67-0cc6-447b-bd9f-342d60833746'`

x1h0:
So far hooks are working now. It seems there is just a race left.

My teardown hook runs on PostDisconnect, which should be the right phase, but it still races: disconnect starts the cancel + close on a detached, un-awaited thread and returns immediately, so PostDisconnect fires while the driver is still opening the KILL connection over the tunnel. My hook then closes the SSH forward and the cancel dies. So PostDisconnect doesn't actually run after the connection is fully closed.

x1h0:
Thanks for the fix. No further problems so far :tada:

0xErwin1:
Glad it's all working now. Thanks for the detailed report and for testing each fix along the way, the teardown race in particular would have been hard to pin down without your logs. Merging.

PR #302, opened 2026-07-16
fix(mcp): make dbflux mcp start and expose its tool catalog

x1h0:
## Summary

Fixes three startup bugs that made the standalone MCP server (`dbflux mcp --client-id <id>`) unusable: a table-name
mismatch in the governance storage layer, a panic while installing the audit bridge sink, and an empty `tools/list`
response caused by an rmcp macro default change. Adds regression tests for the storage and tool-router fixes.

## What does this resolve?

No linked issue; found while setting up the MCP server locally. It failed in three stages, each fix uncovering the next:

1. Startup aborted with `no such table: cfg_connection_profile_governance_binding_roles` for any profile with a governance
binding.
2. Once past that, it panicked with `Cannot block the current thread from within a runtime` at `lib.rs:54`.
3. Once running, clients connected but `tools/list` returned an empty list.

## How was this solved?

- **Storage:** The initial migration creates the binding role/policy tables as `cfg_connection_profile_gov_binding_*`, but
the repositories queried the full-word `cfg_connection_profile_governance_binding_*` names. This did not only break MCP startup: the main app uses the same repositories, so per-profile governance binding roles and policies were silently broken there too. The repository SQL now matches the migrated schema. No data migration is needed: the mismatched tables were never writable, so they are empty everywhere.
- **Startup panic:** `run_mcp_server` is async but acquired the runtime lock via `blocking_read()`, which tokio forbids on
runtime worker threads. Replaced with `read().await`.
- **Empty tool list:** rmcp's `#[tool_handler]` macro defaults to `Self::tool_router()`, the static router of its own impl block, which has no tools. The combined router built in `DbFluxServer::new()` was never served, broken since the tool sub-router split in `3260f8b8`. The router field is now passed explicitly via `#[tool_handler(router = self.tool_router)]`.

## Where was this tested?

- [x] Automated tests
- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] X11
- [ ] Wayland
- [ ] Other:

## Checklist

## Labels to apply

`mcp:bug`, `mcp`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,14 @@ All notable changes to DBFlux will be documented in this file.
 
 ### Fixed
 
+* **Standalone MCP server failed to start and listed no tools** — The
+  governance binding role/policy repositories queried table names that did not
+  match the migrated schema, so `dbflux mcp` aborted with "no such table" for
+  any profile with a governance binding (the same mismatch also broke saving
+  these bindings in the main app). Startup then panicked on a blocking lock
+  read inside the async runtime, and `tools/list` returned an empty catalog
+  because the tool handler served an empty router instead of the combined one.
+
 * **SSH tunnel / proxy list icon vanishing on long hosts** — The globe icon in
   the Settings SSH tunnels and proxies lists could be squeezed to zero width by
   long, unbreakable hostnames (e.g. EC2 `ec2-…compute.amazonaws.com` addresses),
--- a/crates/dbflux_mcp_server/src/lib.rs
+++ b/crates/dbflux_mcp_server/src/lib.rs
@@ -51,7 +51,7 @@ pub async fn run_mcp_server(args: McpServerArgs) -> anyhow::Result<()> {
     // to the audit database.  The clone shares the underlying SQLite
     // connection with the service held by McpRuntime.
     if let Some(ref handle) = bridge_handle {
-        let audit_clone = state.runtime.blocking_read().audit_service().clone();
+        let audit_clone = state.runtime.read().await.audit_service().clone();
         if let Err(err) = handle.install_sink(Arc::new(audit_clone)) {
             log::warn!("Failed to install audit bridge sink: {err}");
         }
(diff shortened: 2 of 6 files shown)


0xErwin1 (approved):
Nice find, and the diagnosis holds up under checking. I verified the branch locally: `cargo check -p dbflux_mcp_server -p dbflux_storage` is clean, and all three tests pass. I also confirmed the rmcp part against the 1.7.0 source, since it was the one claim resting on third-party macro behavior: `#[tool_handler(router = self.tool_router)]` is documented, first-class syntax (`rmcp-macros` has a "Custom router expression" section with that exact example, and its own test suite uses it whenever the router lives in an instance field), and bare `#[tool_handler]` does default to `Self::tool_router()`. Your explanation is right down to the detail.

Two small things before merge:

1. **Drop the `#[allow(dead_code)]` and its comment on `server.rs:373`.** After this fix the field is genuinely read, so the comment ("held to keep tool routes alive") is now false, and the `allow` is exactly what hid the bug in the first place: it suppressed the warning that nothing was reading the combined router. Removing it is what stops the regression from coming back silently.

2. **Add a `CHANGELOG.md` entry under `## [Unreleased]`.** This is user-visible: the MCP server did not start.

One note, non-blocking: the storage fix is bigger than the title suggests. `config_loader.rs:439` and `:446` use the same repositories, so per-profile governance binding roles and policies were broken in the main app too, not just on `dbflux mcp` startup. Worth reflecting in the PR body so the fix is findable later.

The tests are well chosen, by the way. Running the repository SQL against the real migrated schema is precisely the gap that let the table-name mismatch through, and asserting exposure *through* the `ServerHandler` rather than just checking the router is populated is the right assertion for that bug.

x1h0:
Thanks for the review! All points are done: allow dead_code removed, changelog entry added, and the PR body now mentions the main-app impact.

0xErwin1:
Both points addressed exactly, thanks. Nice touch folding the wider `config_loader` impact into the changelog entry, that is the part someone will actually search for later.

The only red check is `Driver Live Integration`, and it is an unrelated flake: `mssql_ssl_mode_on_trusts_self_signed` never ran, it panicked at container startup with `WaitLog(EndOfStream(...))` while SQL Server was still upgrading the master DB. The same job passed on `main` a few hours earlier and this PR does not touch the MSSQL driver. Merging over it.

Thanks for chasing all three bugs down to the root instead of stopping at the first one. The `tool_handler` router default was a genuinely subtle one to catch.

PR #305, opened 2026-07-17
feat(ui): context-aware SQL autocomplete in the query editor

x1h0:
## Summary

Context-aware SQL completion for the query editor. Suggestions now follow where the cursor sits in the statement instead of listing every identifier the connection knows: tables and views after `FROM`/`JOIN`, columns scoped to the tables actually referenced in the statement (even with no prefix typed), relation aliases, CTE names, and `SELECT` output aliases where the engine resolves them (`GROUP BY`/`ORDER BY`/`HAVING`). It is built on the same tree-sitter grammar the editor diagnostics already use, so it holds up on half-typed queries, and it prefetches column metadata in the background so you no longer have to expand a table in the sidebar first.

The branch also carries three smaller fixes that surfaced while building this (drift fingerprint, drift preflight database, and an on-connect snapshot hydration that keeps completion warm after a restart).

## What does this resolve?

No linked issue. This is a proactive improvement to the editor autocomplete.

## How was this solved?

Split into six self-contained changes:

1. **Drift fingerprint:** Foreign keys are hashed in sorted order in the drift fingerprint (both the session and stable variants), matching how `diff_table_info` already compares them. A fingerprint mismatch that produces no reportable diff is now treated as no drift, so a mere FK reorder no longer opens an empty drift modal. This pre-existing bug became visible once completion started populating the table-details cache on its own.
2. **Statement ranges:** `split_sql_statements` was reworked onto a byte-range scanner, and `QueryLanguage::statement_bounds_at` was added so callers can locate the statement under a cursor offset. Splitting behavior is unchanged and still covered by the existing tests.
3. **Context engine:** `SqlContextEngine` classifies the completion slot under the cursor (table vs column position, including the clause) and collects the visible scope (relations with aliases, CTE names, `SELECT` aliases), subquery-aware. Half-typed SQL parses into `ERROR` nodes, so classification recovers through the anchor token's keyword and a backward scan over the leaves before the cursor. Unclassifiable positions report `None` and callers keep a heuristic fallback.
4. **Completion provider:** The provider consumes the engine: scoped columns, alias items, CTE/table/view suggestions, and `sort_text` ranking so context items sort above keywords. Referenced tables prefetch their column details in the background, and lazy-per-database drivers prefetch the selected database's table listing. All sources are scoped to the document's selected database so a tab does not leak tables from other databases. Whitespace after a clause keyword opens the menu, guarded so it never opens without context-relevant items, and deletions re-query the menu so it does not go stale next to removed text.
5. **Drift preflight database:** The schema-drift preflight now resolves its database from the editor selection first, and skips entirely when no database is known on lazy-per-database drivers, instead of falling back to a fabricated `default` that reached real introspection queries as `` `default`.`table` `` and failed with permission errors.
6. **Snapshot hydration:** After the on-connect auto-capture, the session table-details cache is hydrated from the most recent persisted deep snapshot, restricted to tables still present in the live listing, so completion and detail views start warm without driver round trips after a restart. Stale columns are caught by the existing drift check at query time.

Design notes:
- The engine lives in `dbflux_core::query` next to the other language services and reuses the grammar and `extract_referenced_tables` that already existed, rather than adding a parser.
- No per-driver branching. Routing keys off `QueryLanguage`/`SchemaLoadingStrategy` and the generic completion metadata, so every relational driver benefits and non-SQL dialects fall through unchanged.

## Validation

Commands (all pass on this branch):

```
cargo fmt --all -- --check
cargo clippy --workspace -- -D warnings
cargo test --workspace
```

New unit coverage: cursor classification, scope collection, ranking, alias resolution across statements, whitespace-trigger noise guards, the trailing-comma column fallback, and a no-panic test over multi-byte input at every byte offset (editor offsets are floored to a char boundary before slicing). Driver-level `table_details` behavior stays covered by the existing suites.

Manual scenarios in the running app:
- Completion after `FROM`, `WHERE`, `JOIN ... ON`, `GROUP BY`
- Alias qualifiers (`u.` resolving to the aliased table)
- Comma-space in a `SELECT` list once the statement has table context (e.g. `SELECT a, | FROM t`)
- Database switching via the toolbar dropdown
- Deletion re-opening the menu at the new position
- A connection with many databases

## Where was this tested?

- [x] Automated tests
- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] X11
- [x] Wayland
- [ ] Other:

## Checklist

### Known limitations / follow-ups

- The schema-drift preflight still re-fetches `table_details` for referenced tables on every query run. A follow-up could mark cache entries verified after a live fetch and only re-check hydrated snapshot entries once (invalidating on in-app DDL), so repeated runs of the same query stop paying for round trips.
- On-connect shallow auto-captures can push a deep snapshot out of the retention window (shallow and deep fingerprints of the same structure differ), which would cost the hydration warm-start on the next connect. Worth deciding whether deep snapshots should be retention-exempt.
- Snapshot hydration only helps when a deep snapshot exists for the connection (explicit capture or lazy deep back-fill); otherwise completion warms up through the background prefetch as before.
- Non-SQL dialects (Mongo, Redis, etc.) are unchanged and keep their existing keyword completion.

## Labels to apply

`query:feature`, `ui:feature`, `kind:sql`, `query`


0xErwin1 (commented):
Thanks for this, it is really clean work. The commits are atomic and each one carries its own reasoning, the test coverage is thorough (the no-panic-at-any-byte-offset test and the foreign-dialect fixtures are exactly the right instinct), and the driver-agnostic routing via `QueryLanguage` / `SchemaLoadingStrategy` keeps the UI properly decoupled.

I ran the full build and the targeted suites on the branch: `cargo check --workspace` is clean (0 warnings), clippy passes with `-D warnings`, and all 110 tests across `sql_context` / `fingerprint` / `drift` / `statement_bounds` / `completion` / hydration are green.

**Verdict: mergeable.** No correctness or security blockers. I traced the risky paths (the re-entrant deletion re-query, multi-byte offset math, FK fingerprint sorting, the preflight early return, and the hydration race) and they all hold up. I would like two small things before merge, the rest is fast-follow or nits.

## Please address before merge

**1. A transient prefetch failure blocks retry for the whole session** (`prefetch_sql_table_details`)

The dedup key is inserted into `prefetched_tables` before the fetch resolves. On a disconnect mid-typing or a transient introspection error (`prepare_fetch_table_details` returns `Err("Profile not connected")`), the key stays committed, so that table shows no columns for `table.` / `alias.` completion for the rest of the session (the provider is not re-created on reconnect of the same connection/database, only on a database or language change). There is graceful degradation via the heuristic fallback and sidebar-expand, but no retry.

Minimal fix: drop the key from the set on the `Err(_) => continue` path and on the spawned task's `Err` branch, so only successful or already-cached keys stay deduped.

**2. Missing CHANGELOG entry**

This is a user-visible feature and the repo tracks `## [Unreleased]`. Please add a line under `### Added`.

## Fast-follow (not blocking)

**3. `analyze` reparses the whole buffer per keystroke** (`sql_context.rs`)

`parser.parse(source, None)` means no incremental parsing, and `collect_leaves` walks the entire tree on each backscan, so the per-keystroke cost is O(total buffer) on the UI thread even though only the statement under the cursor is needed. `statement_bounds_at` already exists in this same PR and could scope the parse. Negligible for small statements, but visible lag on a large multi-statement script (made more frequent by the new whitespace trigger). Good candidate for a follow-up issue.

## Nits

- `sql_completion_items_with_context` is around 172 lines, past the project's ~100-line design-smell line. The three match arms (`TableRef` / `ColumnRef` / `None`) are each self-contained and could be extracted into helpers that push into `&mut SqlItemSink`, leaving the top function as a dispatcher.
- The "schema-qualified key, then bare-name fallback" logic is duplicated across three sites (the dot-qualifier arm, `scoped_columns`, and `columns_from_referenced_tables`). A single `columns_with_bare_fallback` helper would keep the three copies from drifting.
- One-time unthrottled burst of N concurrent introspection tasks on a wide multi-JOIN (bounded and one-shot per provider, so low priority). A small concurrency gate would smooth the spike.
- The prefetch database-scope filter uses `is_none_or` (None means all databases) while the display path uses `map_or(!lazy_per_database)` (None means nothing on lazy drivers). On a lazy-per-database driver with no resolvable database, prefetch fetches details the menu then filters out. Wasteful, not incorrect.
- Non-backspace length shrinks in SQL mode (accepting a completion shorter than the replaced prefix, or deleting a selection) can trigger the menu reopen. Minor UX surprise, not a correctness issue.
- The shared generation counter is `query_generation` in the provider but `completion_query_generation` in `EditorState`. Aligning the names helps grepping both ends.

Again, nice work. Sort out 1 and 2 and this is good to go.

x1h0:
thanks for the review! pushed the fixes.

both blockers are done (prefetch key only sticks once the fetch spawns and gets dropped on failure now, changelog added).
also grabbed the nits: split the big match fn, deduped the bare-name fallback, aligned the db-scope filter and the generation counter name.

went ahead and did the reparse one inline too since it was small, analyze scopes to the statement under the cursor now instead of the whole buffer (also stops it anchoring across statements). that part wasn't in what you reviewed though, so give it another look.

left two for later: the concurrency gate and the non-backspace reopen, which really needs gpui-component to expose completion_inserting to fix cleanly.

0xErwin1:
Thanks for turning these around so fast, and for going past the two must-fixes to also scope the parse to the statement and collapse the duplicated fallback. Verified on the latest head: build is clean, clippy passes, and the targeted suites are green (112/112). Merging now.

PR #312, opened 2026-07-30
fix(ui): browse per-database schemas with isolated tree state

### LINKED ISSUE #311: [feature] Schemas for MSSQL Database
### Problem

SQL Server Databases are organized in Schemas. So, I think it'd be great if the sidebar tree would also be organized in schemas.

### Proposed solution

Load schemas per database even if lazy loading is enabled.

### Alternatives considered

_No response_

### Primary area

_No response_

### Labels I plan to apply

- [ ] Kind label (e.g. `driver:feature`, `ui:feature`, `query:feature`, `mcp:feature`, `rpc:feature`, …)

reenigneEsrever92:
## Summary

Fix schema browsing for lazy per-database connections, including SQL Server. Group fetched tables and views by their actual schema instead of using the initial connection snapshot. Database-qualified schema and folder IDs keep expansion state independent when databases share a schema such as `dbo`.

Resolves #311.

## Changes

- Integrated current `main`, preserving the contributor's original commit.
- Removed the redundant eager MSSQL schema fetch and its obsolete `TableInfo` initializer. The driver now matches `main`; there are no added connect-time queries or silently discarded errors.
- Kept grouping in the generic sidebar, with selected-database routing/cache identity, database-named single-schema flat layouts, views-only schemas, and empty/schema-less content.
- Extended `Schema`, `TablesFolder`, and `ViewsFolder` IDs with optional database identity. Legacy unqualified IDs still parse and round-trip unchanged.
- Added resolver-level regressions, real expansion-override isolation across two databases sharing `dbo`, and legacy/qualified ID round-trip coverage.

## Validation

Passed locally on Linux:

- `cargo test -p dbflux_core --locked node_id` (33 passed)
- `cargo test -p dbflux_ui_sidebar --locked` (210 passed)
- `cargo check -p dbflux_driver_mssql --locked`
- `cargo check --workspace --locked --features "sqlite,postgres,mysql,mssql,mongodb,redis,dynamodb,cloudwatch,influxdb,clickhouse,turso,aws"`
- `cargo clippy --workspace --locked --features "sqlite,postgres,mysql,mssql,mongodb,redis,dynamodb,cloudwatch,influxdb,clickhouse,turso,aws" -- -D warnings`
- `git diff --check`

An independent verification pass reviewed the final implementation and ran these checks. Sidebar tests emit an existing test-only `saved_charts_folder_id` dead-code warning; the Clippy command above does not check all test targets.

Remote CI must pass before merge. Live SQL Server/MySQL browsing and interactive UI validation have not been repeated for the revised implementation. No historical RED evidence is claimed for the final ID correction. Native review tooling was unavailable; no native review closure is claimed.

## Review scope

The maintainer approved `size:exception` for the cohesive schema-grouping and database-identity correction with regression coverage. Final diff against the integrated `main`: 828 changed lines across five files. The fix commit uses a user-visible Conventional Commit message for the generated changelog; `CHANGELOG.md` is not hand-edited.

### DIFF
--- a/crates/dbflux_driver_mssql/src/driver.rs
+++ b/crates/dbflux_driver_mssql/src/driver.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::collections::{BTreeMap, BTreeSet, HashMap};
 use std::sync::LazyLock;
 use std::sync::atomic::{AtomicBool, AtomicI32, Ordering};
 use std::sync::{Arc, Mutex};
@@ -1850,10 +1850,15 @@ impl Connection for MssqlConnection {
             total_start.elapsed().as_secs_f64() * 1000.0
         );
 
+        let schemas = match &current_database {
+            Some(db) => self.fetch_db_schemas(db).unwrap_or_default(),
+            None => Vec::new(),
+        };
+
         Ok(SchemaSnapshot::relational(RelationalSchema {
             databases,
             current_database,
-            schemas: Vec::new(),
+            schemas,
             tables: Vec::new(),
             views: Vec::new(),
         }))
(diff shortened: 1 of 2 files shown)


0xErwin1:
Thanks for taking this one on, the schema nodes are a real gap for MSSQL. A few things to sort out before this can land.

**1. It doesn't compile against current `main`.** The branch is based on `d3359961`, and `TableInfo` gained a `storage_hints` field in #309. Merging `main` in and running `cargo check -p dbflux_driver_mssql` gives:

```
error[E0063]: missing field `storage_hints` in initializer of `TableInfo`
    --> crates/dbflux_driver_mssql/src/driver.rs:3126:23
```

Rebasing on `main` should be enough to surface it locally.

**2. The sidebar branch reads the wrong schema source.** MSSQL is `LazyPerDatabase`, so the new `else if !schema.schemas().is_empty()` arm sits in the lazy path where `db_schema` (the per-database snapshot just fetched via `schema_for_database`) is already in hand, but it passes the connection-level `schema` instead. That snapshot only ever holds the schemas of the database that was current at connect time: `FetchDatabaseSchemaParams::execute` returns a `DbSchemaInfo`, and `apply_database_refresh_outcome` stores it under `database_schemas` with `schema: None`, so the connection snapshot is never refreshed.

Practical effect on a server with more than one database: expand `AppDb` and you get the *initial* database's schemas and tables rendered under it, with `target_database = Some("AppDb")`, so clicking a table opens a path that may not exist. The correctly fetched `db_schema` is dropped. It looks right in single-database testing, which is probably why it passed the manual check.

Since `schema_for_database` already tags every `TableInfo` / `ViewInfo` with its schema name, my suggestion is to keep the fix entirely in `build_db_schema_content` and group that single `DbSchemaInfo`'s tables and views by their `schema` field into schema nodes. That stays driver-agnostic, fixes it for every `LazyPerDatabase` driver at once, and makes the `driver.rs` change unnecessary.

**3. `fetch_db_schemas(db).unwrap_or_default()` swallows the error.** Per CLAUDE.md a fallible result shouldn't be silently discarded, `.log_err()` is the minimum here. Worth noting this also runs the same two `sys.tables` / `sys.views` queries that `schema_for_database` already runs, so it's two extra round-trips on every connect for data the lazy path fetches anyway.

Happy to look again once the rebase and the schema-source question are settled.

reenigneEsrever92:
I see. Thanks for the response. Grouping as you described under point 2 would be done in the sidebar?

0xErwin1:
Yes, entirely in the sidebar: `crates/dbflux_ui_sidebar/src/tree_builder.rs`, no driver or core change needed.

Concretely, in the lazy branch of `resolve_db_children` (where you currently added the `else if`), take the `db_schema` you already have and split it into one `DbSchemaInfo` per distinct `schema` value across its tables and views, then call `build_db_schema_content` once per group. That is the same shape `build_schema_children` already uses when it iterates `snapshot.schemas()`, so the resulting tree matches what the non-lazy path produces.

Two things worth knowing while you do it:

- It also fixes the cache keys. `build_db_schema_content` builds `SchemaCacheKey::new(database_name, Some(schema_name))` from `db_schema.name`, which today is the *database* name on the lazy path, so types/indexes/FKs/routines are cached under a schema that doesn't exist. Grouping by the real schema makes those keys correct for free.
- Keep the single-group case rendering exactly as it does now. MySQL is also `LazyPerDatabase` and its `fetch_tables_shallow` sets `schema: Some(database)` on every table, so it collapses to one group named after the database and comes out unchanged.

A unit test next to `build_db_schema_content_uses_per_table_schema_when_present` covering the multi-schema case would be a good place to pin the behavior.

ignaciohoulak:
Hi @reenigneEsrever92, are you still planning to continue with this PR? It’s been about a month since our last exchange. Let me know if anything is blocking you or if the sidebar grouping approach needs clarification. If you no longer have time for it, that’s fine too, just let me know so I can plan the next steps.

0xErwin1:
Hi @reenigneEsrever92, are you still planning to continue with this PR? It's been about a month since our last exchange. Let me know if anything is blocking you or if the sidebar grouping approach needs clarification. If you no longer have time for it, that's fine too, just let me know so I can plan the next steps.

reenigneEsrever92:
I'll see if I can get something going this weekend 🤞 I swear 😅 if that works for you - great! If not then don't wait for me.

0xErwin1:
Thanks for the update. If you find you can't get to it this weekend, just let me know and I'll take over to get it resolved.

0xErwin1 (approved):
The revised implementation addresses the schema-source and redundant-query findings. Independent local verification also caught and then verified the fix for cross-database expansion collisions, including legacy ID compatibility. Local checks pass; merge remains gated on the required CI checks. Live UI testing has not been repeated.

PR #329, opened 2026-08-13
fix(ssh): restore key-authenticated tunnels on Windows

narma:
## Summary

Restore key-authenticated SSH tunnels on Windows when connecting through modern OpenSSH servers.

## What does this resolve?

Windows builds used libssh2's WinCNG backend, which offered only legacy Diffie-Hellman key exchange methods that do not overlap with current OpenSSH defaults. After switching the backend, private-key authentication could still fail because libssh2's C file API could not reliably open valid Windows key paths.

- No linked issue.

## How was this solved?

- Use libssh2's OpenSSL backend on Windows so Curve25519/ECDH key exchanges are available, while leaving non-Windows dependency behavior unchanged.
- On Windows, read the private key through Rust into a zeroizing `SecretString` and authenticate with `userauth_pubkey_memory`, avoiding the failing C file-path boundary.
- Keep `userauth_pubkey_file` on non-Windows platforms.
- Add an ignored, environment-driven PostgreSQL live test that verifies the complete key-authenticated SSH tunnel path, including ping and query execution.

The Windows backend uses vendored OpenSSL. This adds work to the first cold build and requires Perl at build time, but it adds no runtime dependency. GitHub's Windows runner already includes Perl, and the existing Rust target cache retains dependency build artifacts for later runs.

## Validation

- `cargo clippy --workspace --locked --features "sqlite,postgres,mysql,mssql,mongodb,redis,dynamodb,cloudwatch,influxdb,aws" -- -D warnings`
- `cargo check -p dbflux --no-default-features --features postgres`
- `cargo test -p dbflux_ssh` — 13 passed
- `cargo test -p dbflux_driver_postgres --lib` — 82 passed, 1 pre-existing live test ignored
- `cargo test -p dbflux_driver_postgres --locked --test live_integration -- --ignored` — 14 passed after isolating the opt-in SSH environment test in its own test target
- Targeted `rustfmt --check` for both changed Rust files and `git diff --check`
- Windows Docker E2E with PostgreSQL 17 reachable only through SSH `direct-tcpip`; `ping` and `SELECT 42` passed with:
  - RSA 3072 PEM key
  - Ed25519 OpenSSH key
  - Encrypted Ed25519 OpenSSH key with passphrase
- A full Windows workspace test run was also attempted. It reached `dbflux_components`, where 323/325 tests passed; two pre-existing style guardrail tests failed on unchanged `chart/engine.rs`. The SSH and PostgreSQL test suites were green.

## Where was this tested?

- [x] Automated tests
- [ ] Linux
- [ ] macOS
- [x] Windows
- [ ] X11
- [ ] Wayland
- [x] Other: Docker Desktop Linux containers (OpenSSH and PostgreSQL 17)

## Checklist

- [x] I did not edit `CHANGELOG.md`; the current contribution guide generates user-visible entries from the `fix` commit
- [ ] I applied the appropriate labels — requested below; fork contributors cannot assign labels in upstream

## Labels to apply

- `ssh:bug`
- `ssh`
- `platform:windows`

### DIFF
--- a/crates/dbflux_driver_postgres/tests/live_integration.rs
+++ b/crates/dbflux_driver_postgres/tests/live_integration.rs
@@ -6,15 +6,18 @@
     clippy::result_large_err
 )]
 
+use dbflux_core::secrecy::SecretString;
 use dbflux_core::{
     CollectionRef, ColumnAssignment, ConnectionProfile, DbConfig, DbDriver, DbError,
     DescribeRequest, ExplainRequest, MutationRequest, OrderByColumn, Pagination, QueryRequest,
     RecordIdentity, RowDelete, RowInsert, RowPatch, SchemaLoadingStrategy, SemanticFilter,
-    SemanticRequest, SqlUpdateRequest, SqlUpsertRequest, TableBrowseRequest, TableCountRequest,
-    TableRef, Value, WhereOperator,
+    SemanticRequest, SqlUpdateRequest, SqlUpsertRequest, SshAuthMethod, SshTunnelConfig,
+    TableBrowseRequest, TableCountRequest, TableRef, Value, WhereOperator,
 };
 use dbflux_driver_postgres::PostgresDriver;
 use dbflux_test_support::containers;
+use std::env;
+use std::path::PathBuf;
 use std::time::Duration;
 
 fn connect_postgres(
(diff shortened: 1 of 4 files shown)


0xErwin1 (approved):
Thanks for the clear fix and thorough validation.

PR #332, opened 2026-08-14
fix(drivers): truncate query previews at UTF-8 char boundaries

### LINKED ISSUE #331: [bug] MySQL query preview panics at a multibyte UTF-8 boundary and crashes the app
### DBFlux version

0.7.4

### Operating system

macOS

### OS / distro detail

 macOS 26.6.1 (25G76), MacBookPro17,1, Apple M1, 8 GB RAM

### Architecture

arm64 / aarch64

### Affected driver (if any)

_No response_

### Steps to reproduce

1. Open a working MySQL connection.
2. Open a SQL editor.
3. Paste the following SQL exactly as shown:

```sql
-- 1234567890123456789012345678901234567890123456789012345678901234567890123456中
SELECT 1;
```
4. Execute the query.

The character `中` starts at byte offset 79, so byte offset 80 falls inside its three-byte UTF-8 representation.

### Expected behavior

The query should execute normally. Generating a log preview must not panic when the SQL contains multibyte UTF-8 characters.

### Actual behavior

DBFlux immediately terminates with `SIGABRT`.

This occurred three times with an identical faulting stack:

- 2026-08-13 18:10:28 +0800
- 2026-08-13 18:10:49 +0800
- 2026-08-14 13:04:21 +0800

The local DBFlux SQLite state database passes `PRAGMA quick_check`, so this does not appear to be caused by database corruption.

### Logs / stack traces

```shell
Application Specific Information:
abort() called

Exception Type: EXC_CRASH (SIGABRT)
Termination Reason: Namespace SIGNAL, Code 6, Abort trap: 6
Triggered by Thread: com.apple.root.user-initiated-qos

system_panic: Application panic captured
end byte index 80 is not a char boundary; it is inside '部'
(bytes 79..82) of `<redacted SQL>`
at crates/dbflux_driver_mysql/src/driver.rs:2060:38


The packaged binary reports line 2060. In the v0.7.4 source, the corresponding query-preview code is currently:


let sql_preview = if req.sql.len() > 80 {
    format!("{}...", &req.sql[..80])
} else {
    req.sql.clone()
};


`String::len()` returns the UTF-8 byte length, while `&req.sql[..80]` requires byte 80 to be a valid character boundary. A query containing Chinese characters, emoji, or other multibyte characters will panic whenever the truncation point lands inside a character.

The same code is still present on `main` at the time of reporting.

A possible fix is to truncate by characters, such as with `req.sql.chars().take(80)`, or otherwise move the byte limit to a valid UTF-8 boundary. Regression coverage should include ASCII, Chinese text, emoji, and values immediately around the truncation boundary.
```

### Labels I plan to apply

- [ ] Kind label (e.g. `driver:bug`, `ui:bug`, `query:bug`, `mcp:bug`, `rpc:bug`, …)

x1h0:
## Summary

The query preview in the debug log was sliced at a fixed byte offset (`&req.sql[..80]`), which panics if byte 80 lands inside a multi-byte character. A query with a CJK char at the wrong spot killed the app.

Reported for MySQL, but Postgres and MongoDB had the same slice and crashed on the same query. mssql already cut on a char boundary.

## What does this resolve?

- Resolves #331

## How was this solved?

All three drivers now use `truncate_string_safe` from core, which we already use for history and saved-query previews. Nothing new, just three hand-rolled copies less.
Left mssql alone since it was already correct.

## Validation

Added tests for `truncate_string_safe` (ASCII, the query from the issue, a cut landing exactly on a char boundary, plus a brute-force pass over every length with mixed 2/3/4-byte chars). Checked they actually fail against the old slicing logic first.

## Where was this tested?

- [x] Automated tests
- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] X11
- [ ] Wayland
- [ ] Other:

## Checklist

## Labels to apply

driver:bug, driver, driver:mysql/mariadb, driver:postgres, driver:mongodb

### DIFF
--- a/crates/dbflux_core/src/lib.rs
+++ b/crates/dbflux_core/src/lib.rs
@@ -262,3 +262,56 @@ pub fn truncate_string_safe(s: &str, max_len: usize) -> String {
 
     format!("{}...", &s[..safe_end])
 }
+
+#[cfg(test)]
+mod truncate_tests {
+    use super::truncate_string_safe;
+
+    #[test]
+    fn keeps_short_strings_verbatim() {
+        assert_eq!(truncate_string_safe("SELECT 1;", 80), "SELECT 1;");
+    }
+
+    #[test]
+    fn truncates_ascii_with_ellipsis() {
+        let long = "a".repeat(200);
+        let truncated = truncate_string_safe(&long, 80);
+        assert_eq!(truncated, format!("{}...", "a".repeat(77)));
+    }
+
+    #[test]
+    fn truncates_inside_multibyte_codepoint_without_panicking() {
+        let digits: String = "1234567890".chars().cycle().take(76).collect();
+        let sql = format!("-- {digits}中\nSELECT 1;");
+        assert!(!sql.is_char_boundary(80), "byte 80 must split 中");
+
+        assert_eq!(
+            truncate_string_safe(&sql, 80),
+            format!("-- {}...", &digits[..74])
+        );
+    }
+
+    #[test]
+    fn truncates_between_every_byte_of_a_multibyte_run() {
+        // 4-byte emoji, 3-byte CJK and 2-byte latin-1 mixed, so every possible
+        // cut index lands somewhere inside a codepoint for some `max_len`.
+        let text = "🚀中ä".repeat(40);
+        for max_len in 0..text.len() {
+            let truncated = truncate_string_safe(&text, max_len);
+            assert!(
+                text.starts_with(truncated.trim_end_matches('.')),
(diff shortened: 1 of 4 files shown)


0xErwin1:
Thanks for this. Clean fix: you routed all three drivers through the existing `truncate_string_safe` instead of adding a fourth hand-rolled truncation, and the brute-force test over every cut index is exactly the coverage this needed.

Merging.

0xErwin1 (approved):
Verified the fix and the tests. Good to go.

PR #334, opened 2026-08-17
feat(clickhouse): add HTTP driver support

fabiokr:
## Summary

Adds built-in ClickHouse and ClickHouse Cloud support for database browsing, typed SQL results, visual SELECTs, charts, and MCP access.

The initial scope intentionally excludes structured mutations, transactions, SSH tunneling, query cancellation, and table transfer.

## What does this resolve?

- Resolves #24

## How was this solved?

Introduces a relational HTTP(S) driver with dynamic ClickHouse type decoding and lazy system-catalog discovery. Generic capability, persistence, and MCP seams keep the UI and application workflow driver-agnostic.

Pagination uses ClickHouse's native HTTP `limit` and `offset` settings so ordered queries are not wrapped or rewritten. The HTTP client retains passwords as `SecretString` until constructing Basic Auth. A branded ClickHouse icon is registered through the generic driver icon seam. SQL governance classification changes are intentionally excluded and are proposed separately in #336.

## Validation

- `cargo test --workspace --features clickhouse` (4,887 passed, 224 ignored)
- `cargo clippy --workspace --features clickhouse -- -D warnings`
- `cargo check -p dbflux --no-default-features --features clickhouse,mcp`
- `cargo test --doc --workspace --features clickhouse`
- `cargo test -p dbflux_driver_clickhouse --test live_integration --no-run`
- `cargo fmt --all -- --check`

The Docker-backed ClickHouse suite includes ordered View pagination coverage for ClickHouse/ClickHouse#33289 and is wired into CI with a pinned ClickHouse 25.8 LTS container. It was compiled locally but not executed because the local Docker daemon was unavailable.

## Where was this tested?

- [x] Local development environment
- [x] Automated tests
- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] X11
- [ ] Wayland
- [ ] Other:

## Checklist

- [ ] I verified the change against the affected user flow(s)
- [x] I added or updated tests when needed
- [x] I documented follow-up work or known limitations when applicable
- [x] Changelog entries are generated from the `feat` commit; `CHANGELOG.md` was not edited per repository policy
- [ ] I applied the appropriate labels

## Labels to apply

- `driver:feature`
- `driver`
- `query`
- `mcp`
- `kind:sql`

PR #336, opened 2026-08-17
fix(governance): classify SQL administration operations

fabiokr:
## Summary

Ensures administrative and disruptive SQL operations receive appropriate governance instead of falling back to generic write policies.

Administrative operations such as `SYSTEM`, `KILL`, `OPTIMIZE`, backup, and restore now require admin governance. Disruptive metadata operations such as `DETACH`, `RENAME`, and `EXCHANGE` require destructive governance.

Extracted from the review of #334 so the policy change can be evaluated independently across all SQL drivers.

## Validation

- `cargo test -p dbflux_core query::safety`
- `cargo test --workspace` (4,858 passed, 220 ignored)
- `cargo clippy --workspace -- -D warnings`

## Checklist

- [x] The change is limited to generic SQL governance
- [x] `CHANGELOG.md` was not edited per repository policy


0xErwin1 (approved):
Thanks for splitting this out. Destructive for DETACH, RENAME and EXCHANGE is the right call.

PR #477, opened 2026-08-28
Add zh_Hans translations for `docs/USAGE.md` and keep the `schema` not translated in `locales/zh_Hans.yml`

leic4u:
Translating the documentation was a bit challenging because I needed to maintain consistency with the translation in the GUI, requiring constant verification of both versions.

Furthermore, since some technical terms require readable localization, I also needed to consult the official driver documentation.

Fortunately, AI agents helped me a lot, and after a little time, I was able to complete the translation task relatively well.

By the way, do the other docs in the `docs/es/*` need to be translated?


0xErwin1:
The translation itself looks solid, but the site won't render `docs/zh_Hans/USAGE.md` yet. Documentation locales are still hardcoded to English and Spanish in three places:

- `web/src/i18n/index.ts` (the `Locale` type and `LOCALES`)
- `web/src/content.config.ts` (the glob lists `*/docs/es/*.md` explicitly)
- `web/src/data/docs.ts` (`parseDocId` only strips an `es/` prefix, and `esTitles` is Spanish-specific)

That is also why the Web checks are green here: the file is invisible to the build, so `check-links.mjs` never walks it.

Could you open a separate PR that adds `zh_Hans` to the web layer, generalizing `esTitles` into a per-locale map? Once that lands, this page renders and it makes sense to talk about translating the rest.

0xErwin1 (approved):

PR #478, opened 2026-08-28
Adds zh_Hans to the web layer, generalizing esTitles into a per-locale map

leic4u:
Since I'm not a developer, I used GLM-5.3 to help me complete this requirement. 

Please carefully check if the submitted code is usable.

And judging from my limited coding skills, it might be fine 😅


0xErwin1:
Thanks for following up on the web layer, merging this. The #470 registry made this exactly the small change it should be, and the Windows fixes in the scripts are welcome too.

Two notes so you know where things stand:

- Every docs page that has no file under `docs/zh_Hans/` serves the English body with a "not translated yet" notice, never a 404. So the remaining pages (the rest of `docs/` and the driver READMEs) can be translated incrementally, one file per PR, with no code changes needed.
- One small thing we'll fix ourselves in a follow-up: the locale id `zh_Hans` is emitted verbatim into `<html lang>` and the `hreflang` links, and underscores are not valid there (the valid form is `zh-Hans`), so search engines ignore those hints. The registry separates the id from the docs directory precisely for this, so `docs/zh_Hans/` stays as is. No action needed on your side.

leic4u:
All right, I will translate the other docs and repo's README next step in following days. 1 doc, 1 PR.

PR #558, opened 2026-09-07
Standardize the translation of "pre-connect" and so on

leic4u:
Standardize the translation of "pre-connect" and so on


0xErwin1:
Thank you

Issue #540, opened 2026-09-03
[feature] Column header context menu for ordering and filtering on right click


0xErwin1:
### Problem

A column header does exactly one thing today: a left click cycles the sort (`cycle_sort`, `crates/dbflux_components/src/components/data_table/table.rs`). Everything else that belongs to a column — filtering it, clearing its ordering — lives in the cell context menu, mixed in with row and cell actions, where nothing signals that those entries are scoped to the column the pointer happened to be over.

### Proposed solution

Right-clicking a column header opens a menu scoped to that column, with the ordering and filtering entries in one flat list rather than nested submenus.

A left click keeps cycling the sort, unchanged. The menu is an addition, and it must not open on left click: sorting by clicking a header is the fastest interaction in the grid, and spending it on opening a menu makes the common case slower to serve the rare one.

### Alternatives considered

A dropdown caret always visible in each header. Rejected: it spends horizontal space in every header for an action taken rarely, and the header already has to fit the PK/FK badges and the type label before it starts truncating the column name.

Leaving these entries in the cell context menu. That keeps column-scoped actions in a menu whose scope is the row, which is the reason they are hard to find.

### Primary area

ui

### Labels I plan to apply

- [x] Kind label (e.g. `driver:feature`, `ui:feature`, `query:feature`, `mcp:feature`, `rpc:feature`, …)
- [ ] Driver label if applicable (`driver:postgres`, `driver:mongodb`, …)
- [ ] Platform label if platform-specific (`platform:linux`, `platform:macos`, `platform:windows`)
- [ ] Data-model label if relevant (`kind:sql`, `kind:document`, `kind:kv`, `kind:log`)

Issue #538, opened 2026-09-03
[feature] Value panel: inspect one cell as JSON, XML, or text in the inspector rail


0xErwin1:
### Problem

A cell holding a JSON document, an XML payload, or a paragraph of prose is unreadable in a grid cell. The grid shows the first characters and truncates, and the only ways out today are widening the column until it swallows the viewport, or copying the value into another tool to read it. Neither lets the value be edited where it is.

### Proposed solution

`View Value` in the cell context menu, plus a keybinding, opens the active cell in the workspace inspector rail.

The panel offers a `JSON` / `XML` / `Text` selector, pretty-print, compact, and word wrap. The format is detected from the value, but only when the value actually parses: a text column whose content happens to start with `{` must open as `Text`, not as JSON showing a parse error the user cannot act on.

Two behaviors that decide whether this is usable:

- Saving from the panel commits the row, not just stages the change. A user editing a 4 KB JSON document in the panel is done with that row when they save it.
- The panel follows the grid cursor, except while it holds an unsaved edit. Following there would silently discard what the user typed.

Two implementation constraints worth pinning now:

- JSON should delegate to the existing `json_editor_view`, so the value panel and the cell-editor modal agree on what valid JSON is instead of disagreeing at the margins.
- XML should be re-emitted from a parsed event stream with indentation, not reformatted as text. Reformatting by hand loses comments and CDATA, and mangles attribute values containing `>`.

### Alternatives considered

Reuse the existing cell-editor modal. A modal covers the grid and cannot follow the cursor, so reading down a column of JSON values means opening and closing it once per row.

Copy to clipboard and read the value elsewhere. That is the current workaround, and having to leave the application to read one cell is what makes this worth building.

### Primary area

ui

### Labels I plan to apply

- [x] Kind label (e.g. `driver:feature`, `ui:feature`, `query:feature`, `mcp:feature`, `rpc:feature`, …)
- [ ] Driver label if applicable (`driver:postgres`, `driver:mongodb`, …)
- [ ] Platform label if platform-specific (`platform:linux`, `platform:macos`, `platform:windows`)
- [ ] Data-model label if relevant (`kind:sql`, `kind:document`, `kind:kv`, `kind:log`)

PR #570, opened 2026-09-09
feat(results): add a record view that transposes the active row

### LINKED ISSUE #537: [feature] Record view: transpose the active row into an editable name/value list
### Problem

A row that is wider than the viewport cannot be read in the grid. Horizontal scrolling moves the column names out of view, so by the time a far-right column is visible there is nothing left to tell you what it is. The same happens to any row with a long text, JSON, or XML column: the cell shows the first few characters and the rest is unreachable without leaving the grid.

`Inspect Row` in the cell context menu is the only current way to see a whole row, and it is a separate read-only surface rather than a way of looking at the result you already have open.

### Proposed solution

A `Grid` / `Record` toggle in the result status bar, plus a keybinding, switches the result between the two layouts.

Record mode transposes the result: the active row fills the whole result area as `Name` / `Value` pairs, with a header naming the position in the result (`Row 3 of 240`).

It has to be an editor, not a read-only presentation. Every field goes through the same `start_editing` / `EditBuffer` path as a grid cell, so dirty marks, `Save Row`, and revert behave identically in both modes — a user who edits in record mode should not have to switch back to the grid to commit.

Navigation transposes with the layout: up/down walk the fields of the current row, left/right walk rows. The transposition belongs in `DataTableState`, not in the render code, so the component's own actions and the workspace keymap resolve direction through one rule instead of two that can drift.

### Alternatives considered

Grow `Inspect Row` into a full modal. Rejected: it is a read-only detour that cannot edit, and it takes the user out of the result to read one row of it.

Word wrap or auto-sized columns in the grid. That helps one long cell and does nothing for a row with thirty columns, which is the case that actually forces horizontal scrolling.

### Primary area

ui

### Labels I plan to apply

ericphamm:
## Summary

Adds a record view to the results grid: `i`, or the Grid / Record toggle in the result status bar, transposes the active row into a Name / Value list that fills the result area, headed by its position in the result (Row 3 of 240). It is an editor, not a read-only surface — every field goes through the same `start_editing` / `EditBuffer` path as a grid cell, so dirty marks, Save Row and revert behave identically in both layouts.

Navigation transposes with the layout: up/down walk the fields of the row, left/right walk rows. The transposition lives in `DataTableState` (`effective_direction` / `effective_edge`), so the component's own actions and the workspace keymap resolve direction through one rule.

Second commit: a click in the empty space below the last field commits an open inline edit, the same way selecting another field does, instead of leaving the editor open on a field the user has left.

## What does this resolve?

- Resolves #537

## How was this solved?

- `DataTableState` gains a `record_mode` flag, a dedicated scroll handle for the field list, and the direction/edge transposition. `scroll_to_cell` scrolls the field list in record mode.
- `DataTable::render` renders `record::render_record` in place of header + body when the flag is set; the grid-only scroll chrome (phantom horizontal scroller, scrollbars) is skipped in record mode because the field list owns its own vertical axis.
- `DataGridPanel` owns the presentation flag (`chrome.record_mode`) and re-applies it after every `rebuild_table`, since a refresh or requery creates a fresh `DataTableState`. Grouped results refuse record mode: an aggregate has no addressable source row.
- The status bar shows the Grid / Record toggle only where the grid itself is on screen (not for charts, the document tree, or grouped aggregates).
- Keymap: `i` in the Results layer → `Command::ToggleRecordView`, routed to the active document.
- Rebased on current `main`, so this sits on top of #541: the editor's Blur/refocus handling is yours, unchanged; this PR adds no second mechanism for it.

Not in this PR (separate PRs, stacked on this one): the JSON/XML value panel (#538) and the column header menu (#540).

## Validation

- `cargo test -p dbflux_components -p dbflux_ui_document -p dbflux_ui_base -p dbflux_ui_windows` pass, including four new tests in `data_table/state.rs`: arrow keys transpose in record mode, Home/End walk fields, leaving record mode restores grid navigation, entering it selects a field when nothing is active.
- Manual, macOS: open a wide table, press `i`, edit a field, Cmd+Enter saves the row; click below the fields with an editor open — the edit is committed and the table keeps focus; toggle back with `i` or the status-bar button; refresh keeps the mode.

## Where was this tested?

- [x] Automated tests
- [ ] Linux
- [x] macOS
- [ ] Windows
- [ ] X11
- [ ] Wayland
- [ ] Other:

## Checklist

## Labels to apply

`ui:feature`, `kind:sql`


crates/dbflux_ui_document/src/data_grid_panel/mod.rs
```diff
@@ -1387,6 +1391,37 @@ impl DataGridPanel {
...
+        self.apply_record_mode(cx);
+        cx.notify();
+    }
+
+    /// Push the panel's record-mode flag onto the current `DataTableState`.
+    fn apply_record_mode(&mut self, cx: &mut Context<Self>) {
+        let record_mode = self.chrome.record_mode;
+        if let Some(table_state) = &self.grid_table.table_state {
+            table_state.update(cx, |state, cx| state.set_record_mode(record_mode, cx));
+        }
```
0xErwin1, line 1462: Please re-check whether record mode is allowed when applying it after a rebuild. If I enable Record on a normal result and then run a grouped query, this reapplies the cached `true` flag without going through the grouped-result guard in `set_record_mode`.

The aggregate stays in Record while its toggle disappears. Please clear the flag for grouped results and add a regression test covering that transition.

crates/dbflux_ui_document/src/data_grid_panel/render.rs
```diff
@@ -3431,6 +3431,16 @@ impl DataGridPanel {
...
 
+        // The record view is a presentation of the data grid, so it is only
+        // offered where the grid itself is on screen: not for charts, not for
+        // the document tree, and not for grouped aggregates that have no
+        // addressable source row.
+        let show_record_toggle = has_data
+            && self.grid_table.table_state.is_some()
+            && current_result_mode != ResultViewMode::Chart
+            && self.view_config.mode != crate::DataViewMode::Document
+            && !self.is_grouped_result();
```
0xErwin1: This condition also enables the Grid/Record toggle in JSON, Text and Raw modes, since it only excludes Chart. Clicking it changes the table’s hidden presentation state without changing the visible result; switching back to Table then unexpectedly opens Record.

Please gate the control on the table actually being displayed, and apply the same eligibility check to the keyboard command.

crates/dbflux_components/src/components/data_table/record.rs
```diff
@@ -0,0 +1,448 @@
...
+                // that moves focus back to the table — doing so blurs the
+                // input, and the blur subscription cancels the edit on the
+                // user's first click into the field they just opened.
+                .when(!is_editing, |d| {
+                    d.cursor_pointer()
+                        .on_click(move |event: &ClickEvent, window, cx| {
+                            state_for_click.update(cx, |state, cx| {
+                                state.focus(window, cx);
+                                state.select_cell(coord, cx);
+                            });
```
0xErwin1, line 381: Please preserve Shift-click range selection here. This always calls `select_cell`, which resets the selection anchor, whereas the grid calls `extend_selection` when Shift is held.

Selecting one field and Shift-clicking another therefore leaves only the second field selected. Please mirror the grid’s modifier handling and cover this with a regression test.

0xErwin1 (changes requested):
Thanks for sending these over!

Please drop the changelog entry; we generate it from commits at release time.

One thing to coordinate: we squash-merge, so #571 will need rebasing after #570 lands, and #572 after #571. Keep only each PR’s own changes on top of the updated `main`. There may be conflicts, especially after the fixes here.

0xErwin1 (approved):
Thanks for the fixes! The requested changes are addressed and CI is green.

PR #571, opened 2026-09-09
feat(results): add a value panel for JSON, XML and text cells

### LINKED ISSUE #538: [feature] Value panel: inspect one cell as JSON, XML, or text in the inspector rail

ericphamm:
## Summary

Adds a value panel for one cell: **View Value** in the cell context menu, or `v`, opens the active cell in the workspace inspector rail with a JSON / XML / Text selector, pretty-print, compact and word wrap. The format is detected from the value only when it actually parses — a text column whose content happens to start with `{` opens as Text, not as JSON with a parse error.

Two behaviours from the issue that decide whether this is usable, both implemented:

- **Saving from the panel commits the row**, not just the edit buffer. Save and Revert are enabled only while the panel holds a modification; Cmd+S while the panel's editor has the keyboard saves the value, not the script.
- **The panel follows the grid cursor, except while it holds an unsaved edit**, where following would silently discard what the user typed.

Stacked on #570: this branch contains that PR's commits and adds one on top. Review only the last commit.

## What does this resolve?

- Resolves #538

## How was this solved?

- `dbflux_components::components::value_format`: `detect_format`, `format_value`, `compact_value`, `validate_value`. JSON goes through `serde_json`; XML is re-emitted from a `quick-xml` event stream with indentation, so comments, CDATA and attribute values containing `>` survive — reformatting it as text would mangle them. Eleven unit tests cover detection (including JSON-shaped text that does not parse), formatting round-trips and malformed input.
- `dbflux_ui_document::data_grid_panel::value_panel`: the rail content (`ValuePanelContent`), the editor, the format switch, Save / Revert. `editor_has_focus` lets the panel report `ContextId::TextInput` while it holds the keyboard, so the results keymap does not claim bare letters typed into it.
- `DataGridPanel`: `request_value_panel`, follow-on-selection gated by `value_panel_can_follow`, and `write_cell_value` split out of `handle_cell_editor_save` so the panel can save without being dismissed. Saving writes the cell and then requests the row commit (`request_save_row_at`).
- The inspector rail is shared between tabs: `TabManager` hands rail state over on tab activation (`RailState`), and `TabManager::open` now emits `Activated` as well as `Opened`, which is what made a freshly opened tab inherit stale rail content before.

Known deviation from the issue text: JSON validation uses `serde_json` directly rather than delegating to `json_editor_view`. Happy to switch if you want the two to share one parser; the detection and formatting are isolated in `value_format` so that is a local change.

## Validation

- `cargo test -p dbflux_components -p dbflux_ui_document -p dbflux_ui_base -p dbflux_ui -p dbflux_ui_windows` pass, including the `value_format` tests and a context-menu test that View Value is offered on read-only results only over an actual row.
- Manual, macOS: open a table with a JSON column, press `v`; edit, Cmd+S — the row commits and the "unsaved changes" counter clears; move the cursor with arrows — the panel follows; type in the panel — it stops following until saved or reverted; switch tabs — the rail shows the active tab's cell, or hides when the tab owns nothing.

## Where was this tested?

- [x] Automated tests
- [ ] Linux
- [x] macOS
- [ ] Windows
- [ ] X11
- [ ] Wayland
- [ ] Other:

## Checklist

## Labels to apply

`ui:feature`, `kind:sql`, `kind:document`


0xErwin1:
Both #571 and #572 now have conflicts with main. Could you rebase #571 first, keeping just the value-panel changes and preserving the fixes from #570? Once #571 lands, we can do the same for #572.

PR #572, opened 2026-09-09
feat(results): open a column menu on right-click of the header

### LINKED ISSUE #540: [feature] Column header context menu for ordering and filtering on right click

ericphamm:
## Summary

Right-clicking a column header opens a menu scoped to that column: order ascending / descending, clear ordering, then every filter operator, in one flat list rather than nested submenus. A left click on the header keeps cycling the sort, unchanged — it is the fastest interaction in the grid and is not spent on opening a menu. There is no always-visible caret in the header, for the reason given in the issue: it would cost horizontal space in every header for an action taken rarely.

Stacked on #571: this branch contains the record-view and value-panel commits and adds one on top. Review only the last commit.

## What does this resolve?

- Resolves #540

## How was this solved?

- `DataTableEvent::ContextMenuRequested` gains an `is_column_header` flag. The header cell emits it with `true` on right mouse-down (after focusing the table); cell and empty-body emits pass `false`.
- `DataGridPanel` keeps the flag on `TableContextMenu`, skips the driver-supplied row actions for a column menu (they have no meaning for a column), and renders `render_column_header_menu_items` instead of the cell menu: ordering entries first, then the filter operators for the column, flat.
- The menu reuses the existing order/filter actions (`ContextMenuAction::Order`, `RemoveOrdering`, `FilterByValue`, `FilterIsNull`, …) and their translations; no new strings.

## Validation

- `cargo test -p dbflux_components -p dbflux_ui_document` pass, including the existing context-menu tests.
- Manual, macOS: right-click a header — the column menu opens at the pointer, ordering entries first; left click still cycles the sort; choose a filter operator — the filter bar is populated for that column; the cell context menu is unchanged.

## Where was this tested?

- [x] Automated tests
- [ ] Linux
- [x] macOS
- [ ] Windows
- [ ] X11
- [ ] Wayland
- [ ] Other:

## Checklist

## Labels to apply

`ui:feature`, `kind:sql`


0xErwin1:
Thank you for all your contributions to DBFlux and for the care you put into these improvements. I really appreciate the time and effort you have dedicated to the project.

PR #576, opened 2026-09-10
Add the correct spaces between Chinese and English text

leic4u:
Add the correct spaces between Chinese and English text


0xErwin1:
Thank you

PR #577, opened 2026-09-10
Add zh_Hans translations for docs/DRIVERS.md

leic4u:
Add zh_Hans translations for docs/DRIVERS.md


leic4u:
This document appears to be the last document in the `docs/` directory, but I found 5 other documents in the `docs/es/` directory: `ARCHITECTURE.md`, `CONTRIBUTING.md`, `PRIVACY.md`, `SECURITY.md`, and `TRADEMARK.md`. Should I translate them?

There are also 13 documents in the `docs/es/drivers/` directory. Should I translate them?

0xErwin1:
Yes, those five documents and the driver READMEs should be translated too, but there's no need to include everything in this PR. We can add them gradually in follow-up PRs, at whatever pace works for you. Thanks for helping with the translations!

PR #578, opened 2026-09-10
fix(results): keep the context menu in the window and open submenus on hover

ericphamm:
## Summary

Three problems with the cell context menu, fixed together because they share one placement rule:

- **Cut off at the edges.** The menu was drawn exactly at the click, so near the right or bottom edge of the panel it ran off screen. It is now shifted left or up to fit, and submenus open to the left when there is no room on the right.
- **Left hanging while scrolling.** The menu stayed where it was while the grid scrolled away underneath. A wheel over the open menu now closes it, as native menus do.
- **Submenus needed a click.** Filter, Order, Generate SQL and Copy as SQL open on hover now, and hovering a plain item closes them. Each submenu hangs in an `anchored` frame that snaps back into the window, so a long filter list opened from a row near the bottom is shifted up instead of being cut off.

## What does this resolve?

## How was this solved?

- `DataGridPanel` measures its size alongside its origin (same canvas). `place_context_menu` in `context_menu/mod.rs` is a pure function: click → origin inside the panel plus which side submenus open on. It runs once before the items are built (the horizontal side decides how the submenu rows are rendered) and once after (the vertical shift needs the item count).
- `sections.rs`: `submenu_frame` positions each flyout beside its row on the chosen side and wraps it in `anchored().snap_to_window_with_margin(…)`. The four trigger rows open their submenu from `on_mouse_move`; plain rows close it (`TableContextMenu::close_submenus`). The overlay closes on `on_scroll_wheel` through one `dismiss_context_menu` helper shared with the existing mouse-down paths.
- Six unit tests on `place_context_menu` (fits, right edge, bottom edge, side threshold, narrow panel, unmeasured panel) and one that pins that the menu panel never clips its contents, since the submenus hang outside it.

## Validation

- `cargo test -p dbflux_ui_document -- context_menu`: 42 tests pass.
- Manual, macOS: right-click near the right edge — menu fits, Filter/Order open to the left; near the bottom — menu moves up, and the Filter submenu opened from the last row is shifted up whole; wheel with the menu open closes it; hover across Filter → Order → Copy switches and closes submenus.

## Where was this tested?

- [x] Automated tests
- [ ] Linux
- [x] macOS
- [ ] Windows
- [ ] X11
- [ ] Wayland
- [ ] Other:

## Checklist

## Labels to apply

`ui:bug`

### DIFF
--- a/crates/dbflux_ui_document/src/data_grid_panel/context_menu/mod.rs
+++ b/crates/dbflux_ui_document/src/data_grid_panel/context_menu/mod.rs
@@ -32,6 +32,67 @@ enum FilterBackend {
     Mongo,
 }
 
+/// Gap kept between the context menu and the panel edge.
+const CONTEXT_MENU_EDGE_GAP: Pixels = Spacing::XS;
+
+/// Width of the widest submenu (the filter list). A menu whose right edge
+/// leaves less than this to the panel edge opens its submenus to the left,
+/// provided the left side has the room.
+const SUBMENU_MAX_WIDTH: Pixels = px(280.0);
+
+/// How far a submenu overlaps the menu it hangs off (menu width 180 less the
+/// 172 offset in `sections.rs`), so the room it needs is its width less this.
+const SUBMENU_OVERLAP: Pixels = px(8.0);
+
+/// Where a context menu goes, in panel coordinates.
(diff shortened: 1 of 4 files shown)


0xErwin1:
The tests are failing, and this PR has merge conflicts.

PR #579, opened 2026-09-10
fix(results): let the Export button close its menu

ericphamm:
## Summary

A second click on **Export** did not close its menu. The menu closed on a press outside it (`on_mouse_down_out`, capture phase), and then the same click reached the button, whose handler toggles the menu — so it closed and reopened in one click. A full-panel layer under the menu now takes any press or wheel, closes the menu and stops the event; the button reads as a toggle, and a click anywhere else still dismisses.

## What does this resolve?

## How was this solved?

- `render_export_backdrop` in `data_grid_panel/render.rs`: a deferred, full-size, transparent layer rendered while `export_menu_open`, closing on left/right mouse-down and on wheel, with `stop_propagation` so the press does not reach the Export button beneath it.
- The menu itself moves to deferred priority 2 so it stays above the backdrop; its `on_mouse_down_out` is gone.

## Validation

- Manual, macOS: Export opens; Export again closes; open then click into the grid — closes; open then choose CSV — export runs as before.

## Where was this tested?

- [ ] Automated tests
- [ ] Linux
- [x] macOS
- [ ] Windows
- [ ] X11
- [ ] Wayland
- [ ] Other:

## Checklist

- [ ] I added or updated tests when needed (interaction-only change; no unit seam)

## Labels to apply

`ui:bug`

### DIFF
--- a/crates/dbflux_ui_document/src/data_grid_panel/render.rs
+++ b/crates/dbflux_ui_document/src/data_grid_panel/render.rs
@@ -160,6 +160,9 @@ impl Render for DataGridPanel {
             .when_some(self.context_menu.as_ref(), |d, menu| {
                 d.child(self.render_context_menu(menu, st.is_editable, &st.theme, cx))
             })
+            .when(self.chrome.export_menu_open, |d| {
+                d.child(self.render_export_backdrop(cx))
+            })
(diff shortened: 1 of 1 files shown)


0xErwin1:
The GitHub Actions checks are failing.

PR #580, opened 2026-09-10
fix(results): hold the no-primary-key banner until the key is fetched

ericphamm:
## Summary

Opening a table from the sidebar flashed the "This table has no primary key – editing is disabled" banner for a moment, then hid it. The rows load before the table details that carry the primary key, so the grid was briefly read-only for want of a key it usually has. The panel now records that the details fetch is in flight and shows the banner only once it has answered; a table that really has no key still gets it.

## What does this resolve?

## How was this solved?

- `DataGridPanel::pk_details_pending` is set when `fetch_table_details_for_pk` starts and cleared on both its success and error paths.
- `show_pk_warning` in `render.rs` adds `&& !self.pk_details_pending`.

## Validation

- Manual, macOS: opening several tables from the sidebar no longer flashes the banner; a table without a primary key still shows it once loaded.

## Where was this tested?

- [ ] Automated tests
- [ ] Linux
- [x] macOS
- [ ] Windows
- [ ] X11
- [ ] Wayland
- [ ] Other:

## Checklist

- [ ] I added or updated tests when needed (a timing-only change in an async path)

## Labels to apply

`ui:bug`

### DIFF
--- a/crates/dbflux_ui_document/src/data_grid_panel/mod.rs
+++ b/crates/dbflux_ui_document/src/data_grid_panel/mod.rs
@@ -589,6 +589,10 @@ pub struct DataGridPanel {
     runner: DocumentTaskRunner,
     focus_handle: FocusHandle,
     panel_origin: Point<Pixels>,
+    /// A table-details fetch for the primary key is in flight. Until it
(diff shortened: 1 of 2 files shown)


0xErwin1:
Thank you

PR #617, opened 2026-09-14
fix(ui): make confirm modals keyboard-accessible

xilec:
## Summary

Confirm modals are keyboard-driven now: Enter confirms, Escape cancels,
and focus moves off the SQL editor while the dialog is open. Previously
every confirm dialog in the app was click-only, and typing went into the
editor buffer behind the modal.

## What does this resolve?

- None tracked. Found while testing: the multi-statement script
  confirmation introduced in #83 reuses the dangerous-query modal
  pattern, but the `ConfirmModal` keymap context had no layer in
  `default_keymap()`, so the Cancel/Execute handlers that dangerous
  query already had were unreachable, and the script confirmation was
  never wired into `active_context()`/`dispatch_command` at all.

## How was this solved?

- **Keymap** (`dbflux_ui_base`): add a `confirm_modal` layer binding
  Enter → `Command::Execute` and Escape → `Command::Cancel`. The context
  has no parent, so nothing else resolves while a confirm modal is up.
  This also makes the existing Cancel/Execute handlers reachable in
  other documents (dangerous query, key-value delete, object browser
  confirms).
- **CodeDocument** (`dbflux_ui_document`): `active_context()` returns
  `ContextId::ConfirmModal` for `script_confirm` as it already did for
  `dangerous_query`; `dispatch_command` routes Cancel/Execute to
  `cancel_script_query` / `confirm_script_query`.
- **Focus**: opening either pending modal moves focus from the editor
  input to the document container — without this the focused input
  consumes Enter as a newline before the keymap sees it. Confirming or
  cancelling restores editor focus via the existing `focus()`.
- **UI**: the script-confirm Run button uses the `.primary()` variant,
  matching the `mutation_confirm` / `unsaved_changes` convention
  (destructive confirms stay `.danger()`).

## Validation

- `cargo check --workspace` clean; `cargo clippy -p dbflux_ui_base
  -p dbflux_ui_document -- -D warnings` clean.
- `cargo nextest run -p dbflux_ui_base -p dbflux_ui_document`:
  1381 passed. New test `confirm_modal_layer_binds_enter_and_escape_only`
  pins the layer to exactly Enter/Escape with no fallthrough.
- Manually verified against PostgreSQL: a multi-statement script runs on
  Enter and cancels on Escape, typing no longer edits the buffer behind
  the modal, focus returns to the editor afterwards; dangerous-query
  confirm/cancel is keyboard-driven too.

## Where was this tested?

- [x] Automated tests
- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] X11
- [x] Wayland
- [ ] Other:

## Checklist

      *(no triage access to the base repo — suggesting `ui:bug`)*


0xErwin1:
Please route Enter/Escape to the visible workspace confirmation before dispatching to the sidebar. Tracing the new bindings shows that selecting a disconnected connection, opening its menu with `m`, choosing Delete, and pressing Enter reaches `connect_to_profile`, leaving the dialog open. Escape clears the pending deletion without closing the dialog, so clicking Delete afterwards does nothing. Both keys should resolve and close the modal through its existing outcome handlers, with regression tests for these paths. This needs fixing before merge. I verified the dispatch paths in code, but have not reproduced this in the running UI.

xilec:
Good catch — fixed.

`Workspace::dispatch` now routes Enter/Escape to the visible workspace
confirmation before any other dispatch domain, resolving the modal
through its existing outcome handlers: Enter confirms and closes the
delete-connection dialog, Escape cancels and closes it (no more
dead-clicked Delete after a cleared pending delete). The same routing
covers the unsaved-changes and dashboard/saved-chart delete confirms,
which sat in the same `ConfirmModal` arm but were never routed either.

Two follow-on issues the routing exposed while testing:
- the editor input kept keyboard focus when these dialogs opened, eating
  Enter before the keymap saw it — `CloseCurrentTab` no longer refocuses
  the document when the dirty check opened the confirmation instead;
- the unsaved-changes save used a hard Save-As; it now uses the smart
  save (silent write for file-backed documents, Save As only for
  untitled ones) and completes the interrupted close, so Enter saves and
  closes in one step.

Regression tests cover resolve-and-close for both keys on all four
modals, the precedence order between them, and the no-modal fallthrough.

One note: the test module avoids `use super::*` — combining the parent
glob with `#[gpui::test]` sends the gpui_macros expansion into unbounded
recursion (an empty test body crashes rustc); explicit imports sidestep
it.

0xErwin1 (changes requested):
Blocking: `workspace/mod.rs:541` closes tabs immediately after dispatching `SaveQuery`, but saving is asynchronous and Save As can be cancelled. Close each tab only after a successful save; keep it open on cancellation or write failure. Please add regression tests for both cases.

xilec:
Fixed in 0cbb946. The close intent now travels with the document: `PaneHandle::save_for_close`
arms a close-driven save, the document reports `DocumentEvent::SaveFinished { succeeded }` and
emits `RequestClose` only when that save landed, and the workspace closes the tab on
`RequestClose` only — it keeps no pending-close state. A dismissed Save As, a failed write, or
edits typed while the write was in flight consume the flag (`mem::take`) and leave the tab open
with its changes; a later manual save cannot close it. Regression tests cover both cases
(landed/failed close-driven save) in `dbflux_ui_document/src/code/mod.rs`, plus the
workspace-level close on `RequestClose`.

Also fixed while in this dialog: "Don't save" now discards only the documents the dialog listed
(it used to close every open tab); a document with no save path (grid with staged edits) keeps its
tab and says so; Enter no longer confirms an empty selection; and the Enter/Escape routing from
your earlier comment now runs before every other dispatch domain, including the sidebar guards.

0xErwin1 (approved):
Approved. The confirmation-routing and close-before-save blockers are addressed in 0cbb946, and the 12 focused regression tests pass. I checked Save As cancellation in code but did not exercise the native dialog; the missing cancellation-specific test is non-blocking.

xilec:
The Save As cancellation gap from this review is now covered by the follow-up PR: https://github.com/0xErwin1/dbflux/pull/627

The follow-up adds `SaveTargetOutcome` and a shared `resolve_save_target` helper, with a per-`AppStateEntity` override for tests. It also routes the other save-dialog call sites through the same resolver, removing the duplicated native/fallback branching. The regression test verifies that an untitled dirty buffer stays open and dirty, the close intent is dropped, and a later successful save does not close the tab.

PR #622, opened 2026-09-15
fix(postgres): connect without a password in keyword mode and surface server error text

xilec:
## Summary

Two related fixes on the PostgreSQL keyword-mode (non-URI) connect path:

1. **Empty `password=` key broke parsing.** A profile without a password produced a
   libpq string with an empty `password=` in the middle. libpq skips whitespace after
   `=` and reads an unquoted value up to the next whitespace run, so the next pair was
   consumed as the password value: `dbname=` was lost, the server fell back to the
   default database (the user name), and the connection failed with
   `FATAL: database "<user>" does not exist`.
2. **Connection errors were opaque.** tokio-postgres' `Display` names only the error
   kind, so a server-rejected connection reached the UI as the terse
   `Connection failed: Connection error: db error` — the FATAL text lives in the
   error's cause and was dropped by the formatters.

## What does this resolve?

- Keyword-mode profiles with no password (e.g. `trust`/`peer` auth) could not connect
  at all; URI mode was unaffected.
- All PostgreSQL connection failures looked identical ("db error"), making profiles
  impossible to diagnose from the UI.

- Resolves #

## How was this solved?

Commit 1 — `fix(postgres): omit empty password key from keyword conn string`
- Connection strings are built by a pure `build_keyword_conn_string(host, port, user,
  password: Option<&str>, database)`: the `password` key is omitted entirely when the
  password is `None` or empty, and values are quoted/escaped per libpq rules
  (single-quote values containing whitespace, escape `\` and `'` inside quotes) —
  no escaping was applied before.
- Both the direct and the SSH-tunnel paths pass `Option<&str>` through; no
  `unwrap_or("")` anywhere on the path.

Commit 2 — `fix(drivers): surface postgres server messages in connection errors`
- `PostgresErrorFormatter` and `RedshiftErrorFormatter` flatten the error cause chain
  before formatting connection/URI errors: server-reported errors carry severity,
  message, and SQLSTATE (`FATAL: database "missing_db" does not exist (SQLSTATE 3D000)`),
  io/TLS failures append their underlying cause, which also lets the known-cause
  branches ("does not exist", "Connection refused", …) match the real reason.

Other SQL drivers were checked and are not affected by (1): MySQL, MSSQL, and Redshift
pass credentials through typed config builders; ClickHouse keeps the password as an
`Option` in its HTTP client.

## Validation

- New unit tests: 6 for the keyword-string builder (key omitted for `None`/empty,
  present when set, quoting/escaping, two parse round-trips through the real
  `postgres::Config` parser), 3 for error flattening.
- New live tests (Docker, `#[ignore]`d as the rest of the suite): keyword-mode connect
  without a password lands on the right database; a missing database surfaces the
  server text + SQLSTATE; a closed port surfaces the io cause.
- `psql` A/B on the repro container: `host=… user=testuser dbname=testdb` connects;
  `user=testuser password= dbname=testdb` fails with
  `FATAL: database "testuser" does not exist`.
- `cargo test -p dbflux_driver_postgres` (103 passed), `cargo test -p dbflux_driver_redshift`
  (102 passed), `cargo clippy --workspace -- -D warnings`, `cargo fmt --all --check`.
- Manually verified in the app (Linux/Wayland session): keyword mode without password,
  keyword + password with spaces/quotes, URI + password, plain-password regression,
  and the detailed error text on a deliberately wrong database name.

## Where was this tested?

- [x] Automated tests
- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] X11
- [ ] Wayland
- [ ] Other:

## Checklist

  (changelog is generated by git-cliff from commit messages; both commits use `fix`)

## Labels to apply

- `driver:bug` (kind)
- `driver:postgres` (driver)


0xErwin1 (approved):
Thank you!

PR #625, opened 2026-09-16
fix(editor): autosave scripts to their real files and close without losing edits

### LINKED ISSUE #621: [bug] Closing a dirty document from the tab bar silently discards unsaved changes
### DBFlux version

0.8.0-dev.0

### Operating system

Linux (Wayland)

### OS / distro detail

NixOS, kernel 6.18, Wayland

### Architecture

amd64 / x86_64

### Affected driver (if any)

None — UI-level (workspace tab closing)

### Steps to reproduce

1. Open a file-backed SQL script and modify it (dirty).
2. Click the `X` on the tab (or middle-click it, or use the tab context menu → Close).
3. The tab closes immediately; the edit is lost. No confirmation appears.

Same for the batch variants: Close Others, Close All, Close Tabs to Left / to Right.

### Expected behavior

Every close path for a dirty document should route through the
unsaved-changes confirmation the way `Ctrl/Cmd+W` does
(`Workspace::close_active_tab` checks `dirty_summaries` and opens
`ModalUnsavedChanges`).

### Actual behavior

Only `Ctrl/Cmd+W` is protected. Every other close path funnels into
`Workspace::close_tab`, which closes unconditionally, so modified
documents are discarded without warning. Silent data loss, no undo.

### Logs / stack traces

Not applicable — no errors logged; the close happens silently.

### Labels I plan to apply

- [x] Kind label (e.g. `driver:bug`, `ui:bug`, `query:bug`, `mcp:bug`, `rpc:bug`, …) — suggesting `ui:bug`

0xErwin1:
## Summary

- A script open in DBFlux autosaves to its own file, so closing the tab and reopening the script shows the latest text.
- Every close route — the close button, middle-click, the tab context menu, close others/all/to the left/to the right, and `Ctrl+W` — persists pending edits before the tab is removed, and leaves the tab open when the write cannot land.
- Autosave, closing and quitting never write over a file that changed outside DBFlux: the write is refused, the buffer stays dirty, and the refusal is reported. `Ctrl+S` and Save As stay deliberate overwrites.

## What does this resolve?

- Resolves #621

The report was that closing a dirty document from the tab bar silently discarded unsaved changes, while `Ctrl+W` asked first. The direction agreed in the issue discussion was automatic persistence rather than more confirmation dialogs: closing should be recoverable, not destructive. This PR implements that for scripts and makes every close route behave the same way.

## How was this solved?

| Area | Decision |
|---|---|
| Autosave target | A file-backed script is written to its own file on the existing autosave interval. Untitled content keeps a sessions-folder copy. |
| Write safety | Bytes are staged in a private file beside the target and renamed in, so a failed or partial write cannot truncate the original. Permissions survive the replacement, a read-only target is refused, and a symlinked document keeps the link. |
| Conflict protection | An automatic write requires a path-associated baseline: the raw bytes this document last loaded or wrote, paired with their path. A file changed outside DBFlux, a deleted file, an unknown baseline and a path mismatch all refuse the write and keep the buffer dirty. |
| Ordering | One per-document queue serializes autosave, `Ctrl+S`, Save As, close and exit writes. The next write starts only from the previous one's completion, so a save cannot interleave with an autosave. |
| Closing | One funnel for every route. The tab is removed only once the write lands, and it stays open with its buffer when the write is refused. The unsaved-changes dialog no longer appears for code documents, because their file already holds the edits. |
| Quitting | Shutdown flushes pending document edits before cancelling tasks. It never closes tabs, so the session that restores on next start is preserved. |
| Empty scripts | Closing an unedited empty script deletes its file only while the file still holds exactly what DBFlux last wrote. |
| Degraded storage | A scripts folder that cannot be created at startup is reported once with what it means and what to check, instead of only reaching the log. Such queries keep working from the session store. |
| Documentation | The four affected pages and their existing Spanish and Simplified Chinese counterparts were updated in the same PR. |

### Review path

This is one PR with a maintainer-approved `size:exception`, so the commits are the review slices. Each was independently verified before it was committed:

| Commit | Slice |
|---|---|
| `df1ee4c6` | Autosave to the real file, staged replacement, baseline protection |
| `163c8b8f` | Autosave follows the file a Save As retargets |
| `3d81bdda` | One close funnel, conflict-checked close flush |
| `fda7151b` | Exit flush before shutdown cancels tasks |
| `a6135ec7` | Empty-script cleanup keeps files that changed on disk |
| `0e33e443` | Report a scripts folder that cannot be created |
| `c603f274` | Clippy fix for the exit flush loop |
| `a95ab663` | Documentation and translations |

Suggested order for a first pass: `df1ee4c6` (the safety core), then `3d81bdda` (closing), then the rest.

### Intentionally out of scope

- Persistent undo/redo history.
- A recovery-drafts picker: a new query is already a real file under Scripts, so persistence replaces the need for one.
- Migrating legacy pathless session tabs into Scripts files.
- A cross-document registry for two documents saving to the same path.

## Validation

| Command | Result |
|---|---|
| `cargo clippy --workspace -- -D warnings` | clean |
| `cargo nextest run --workspace --test-threads=1` | 6477 passed, 249 skipped |
| `cargo test --doc --workspace` | clean |
| `cargo fmt --all -- --check` | clean |
| `git diff --check` | clean |

Behavioural evidence per slice (each slice observed its failure before the fix):

- Autosave: an injected partial write left the original as `REPLAC` before the staged commit and preserved it after.
- Baseline: an autosave with no loaded baseline created the file before the fix; a recovered shadow differing from disk was granted overwrite authorization before the fix.
- Save As: a debounce armed before a retarget refused as an unknown baseline and stranded the newest text before the fix.
- Closing: a dirty tab was removed before its write landed; two close gestures chained writes forever; a landing `Ctrl+S` silently dropped the close gesture.
- Exit: a pathless document's scratch copy was never written on quit.
- Empty scripts: an empty buffer over a file with foreign content deleted that file before the fix.

### Verification honesty

- The scripts-folder warning slice has no observed RED: a `ScriptsDirectory` failure is not injectable on the surfaces that change touches, so its three tests exercise the drain boundary only. The recording itself is inspection-verified.
- The binary shutdown wiring (the weak workspace handle, its phase placement, and the timeout reporting) is inspection-verified: no unit test covers `run_shutdown_sequence`, and no test drives the flush wait against a document that is really writing.
- `cargo nextest run --workspace` in its default parallel mode fails one unrelated timing test, `palette_filtering_large_dataset_completes_within_budget` (a 50 ms wall-clock budget). It passes serially, its source predates this branch, and no run against the base ref was made, so it is reported as observed and not attributed to this change.

## Where was this tested?

- [x] Automated tests
- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] X11
- [ ] Wayland
- [ ] Other:

No manual GUI pass was performed; the close and autosave flows are covered by tests rather than by hand. Non-Linux behaviour of the staged replacement is unverified.

## Checklist

- [ ] I updated `CHANGELOG.md` under `## [Unreleased]` if this is user-visible — not applicable: CONTRIBUTING.md says the changelog is generated from commit messages with git-cliff and must not be hand-edited.

## Labels to apply

`ui:bug`, `size:exception`

## Known limitations

- There is no power-loss durability guarantee: the staged write is not fsynced, so a crash can still lose the newest bytes.
- Replacing a file changes its inode; a pre-existing hard link keeps the old bytes, and ownership, ACLs and extended attributes are not carried over.
- The destination directory must be writable, so a writable file inside a non-writable directory can no longer be saved.
- Two open documents saving to the same path are not coordinated.
- Every refused autosave debounce reports an error, so a file changed outside DBFlux produces repeated errors while the user keeps typing.
- `is_file_backed_empty` now means "empty and still ours to delete"; the name was kept to hold the diff down.
- The website build and its link checker were not run; no relative links were changed.


xilec:
Post-merge review notes from a read-through of this PR (nothing blocking — the write queue, baseline protection, staged replacement and the close funnel all check out, and `cargo check --workspace` is clean at `23cfcfe`). Two behavioural points worth a follow-up, plus a few hygiene nits.

**1. A file-backed tab without a trustworthy baseline can never be closed, and nothing points at the way out.**
`resolve_close` always queues a conflict-checked `CloseFlush` for file-backed documents (`file_ops.rs:350`). With no baseline — or a baseline belonging to another path — the write refuses as `BaselineUnknown`, the tab stays open, and the next close gesture repeats the cycle, each time with an "Auto-save failed" toast. This state is reachable in production: session restore deliberately leaves the baseline unseeded when the physical read fails or a recovered shadow differs from disk (`actions/query.rs:486`). The Known-limitations entry covers repeated refused autosaves *while typing*, not this close loop. A `suggested_action` on the refusal ("use Save As to keep these edits") would close the gap.

**2. The "Don't save" path is gone for dirty untitled buffers.**
The unsaved-changes dialog is suppressed for every code document (`decides_own_close` keys off `has_close_policy()`, and every `CodeDocument` registers `resolve_close`), so closing a dirty untitled buffer now forces Save As; dismissing it keeps the tab open, and there is no way to close without saving anymore — clearing the buffer leaves it dirty. The rationale in the description ("their file already holds the edits") doesn't hold for untitled buffers: there is no file. If dropping discard is an intentional trade-off, it deserves a line in the description and `docs/USAGE.md`; if not, untitled buffers should keep the dialog.

Minor hygiene, in case any of it is worth an issue:

- Refusals of `CloseFlush` / `ShutdownFlush` report through the `auto_save_failed` toast key, so closing a tab or quitting the app yields "Auto-save failed" (`file_ops.rs:158-230`).
- `mark_clean` still ends with `let _ = std::fs::remove_file(&shadow)` (`code/mod.rs:1409`) — a pre-existing line, but every landed autosave now runs through it, and AGENTS.md forbids the pattern; the new code nearby already does it right with `log::warn!`.
- New blocking fs I/O on the foreground thread: `file_backed_empty_path` reads the file synchronously on every close of an empty tab (`file_ops.rs:275`), and `flush_for_shutdown` rewrites the shadow on every 50 ms shutdown poll (~40 identical writes per document inside the 2 s budget).
- `TabManager::close_others` / `close_all` / `close_to_left` / `close_to_right` / `close_active` still bypass the funnel (`tab_manager.rs:534`); no live callers today, but the first future caller silently reintroduces the regression this PR fixed. `CloseAllTabs` also duplicates the batch loop instead of calling `close_tabs_batch` (`workspace/mod.rs:1222`), and `charts_dashboards.rs:1126` is a second live direct `mgr.close`.
- The exit log prints "All pending document edits flushed" even when the flush was refused or the app context was already gone (`main.rs:541`); the flush timeout is only visible in the log.
- Smaller: hardcoded English strings in the scripts-folder diagnostic (`app_state_entity.rs:60`); a new bootstrap test depends on the real `dirs::data_dir()` (`app_state/mod.rs:3158`); a redundant `CloseFlush` queues behind a clean buffer while an explicit save is in flight; the now-unused `save_script_failed` i18n key; Save As on a clean buffer retargets without `cx.notify()`, so the tab title keeps the old filename.

Line numbers refer to the PR head (`23cfcfe`). Happy to split the minor items into issues if that's easier to track.

xilec:
@0xErwin1, sorry for the late review — by the time I finished reading through, the PR had already merged. I hope the notes above are still useful and help make the app even better. Thanks for your work on this!

0xErwin1:
Thanks for reading through this so carefully, and sorry for the slow reply. Your first note sat for about ten hours before I got to it.

I went back over both behavioural points against current `main` and they hold up. The close loop is reachable exactly as you describe: session restore leaves the baseline unseeded when the file is unreadable or when the recovered shadow differs from disk, `resolve_close` queues the conflict-checked flush without consulting the baseline, and the refusal keeps the tab open behind an "Auto-save failed" toast with no way out. The untitled case is the same root cause from the other side. Every `CodeDocument` registers `resolve_close`, so `has_close_policy()` is true for untitled buffers too, and the dialog that used to offer "Don't save" never appears. I am working on both while this thread continues.

The hygiene items are correct as written, and two of them matter for reasons past tidiness: the refusals reporting through the autosave key make a close or a quit say "Auto-save failed", and `flush_for_shutdown` rewriting the shadow on every 50 ms poll is around 40 synchronous writes per dirty document on the path where the process is trying to finish. One item I could not confirm from reading alone is the redundant `CloseFlush` behind a clean buffer while an explicit save is in flight. If you have a repro, send it and I will fold it in.

Go ahead and open the issue, even though the fixes are already in flight. I would rather the tracking exist and get closed against the pull request than live only in this thread. Splitting the minor items into their own issue works too if that is easier to follow.

xilec:
The repro for the close-flush item is up: #648. My wording in the review was misleading — "a clean buffer" — the buffer is still dirty-flagged at close time (the flag clears only when the write lands), which is probably why it could not be confirmed from reading alone. What actually happens: `resolve_close` queues a `CloseFlush` carrying the exact bytes the in-flight explicit save is already writing, and `execute_write` has no identical-content short-circuit, so the flush re-runs the full staged replacement. The issue includes a deterministic test (passes on current `main`, asserting two `SaveFinished(true)` events for one Ctrl+S plus one close); the same interleaving is already visible in `an_explicit_save_landing_before_the_queued_close_flush_still_closes_the_tab`.

Tracking is filed for everything else from the review as #640–#651 — close them against your fixes as they land. One caveat: the "Save As on a clean buffer keeps the old tab title" item I could not confirm from reading alone and left unfiled; shout if you want it filed anyway.

PR #627, opened 2026-09-17
refactor(ui): unify save-target dialog resolution

xilec:
## Summary

Unify save-destination resolution for save dialogs behind `SaveTargetOutcome` and a shared `resolve_save_target` helper, and add a per-`AppStateEntity` test/embedding override so Save As cancellation can be tested without a native dialog.

Rebased onto `main` after #625 and #630.

## What does this resolve?

- Follow-up to #617: adds the missing cancellation-specific regression test.
- Removes duplicated native-dialog / fallback-export branching from four save-destination call sites.

## How was this solved?

- Added `SaveTargetOutcome` with explicit `Selected`, `Cancelled`, and `Failed` states.
- Added `file_dialog::resolve_save_target(...)`, which owns the native-availability probe, fallback export directory, and cancellation policy.
- Added a per-`AppStateEntity` `SaveTargetProvider` override. Production constructors leave it `None`; tests install it explicitly.
- Routed code Save As, data-grid export, object download, and connection export through the shared resolver.
- Code Save As keeps #625's write queue: the resolver picks the target, and the write still goes through `enqueue_save_as_write`, so retargeting, scratch cleanup and the fallback toast stay where #625 put them.
- Consolidated the Save As tests onto the picker override:
  - `cancelling_save_as_keeps_the_tab_open` — a real `SaveTargetOutcome::Cancelled` leaves an untitled buffer dirty, drops the close intent, creates no file, and a later save still lands without asking to close.
  - `choosing_a_path_in_save_as_writes_and_closes` — a chosen path receives the captured buffer and the interrupted close finishes; it now also asserts that the document is retargeted at that path.
  - `save_as_from_the_toolbar_retargets_without_asking_to_close` — Save As that is not part of a close writes, retargets, and emits no `RequestClose`.
  - #625's `an_untitled_dirty_document_saves_through_save_as_and_keeps_its_buffer_when_dismissed` (which dropped `_pending_save` and called `report_save_outcome(false)` directly) and `save_as_writes_the_captured_content_and_retargets_the_document` (which called `enqueue_save_as_write` itself) are removed; both outcomes are now reached through the seam, and the retarget claim they added is asserted by the tests above.

## Validation

- `cargo fmt --all` — passed.
- `cargo clippy --workspace -- -D warnings` — passed.
- `cargo nextest run -p dbflux_ui_base -p dbflux_ui_document -p dbflux_ui_windows -p dbflux_ui` — 2031 run: 2025 passed, 3 skipped, 6 failed. The six are `dbflux_ui` document-cleanup tests that seed a file under this sandbox's read-only `$HOME` (`ReadOnlyFilesystem` at `actions/documents.rs:753`); they arrived with #625 and are untouched by this diff.
- `cargo nextest run --workspace --no-fail-fast` — 6478 run: 6459 passed, 249 skipped, 19 failed. All 19 are environment-dependent in this sandbox: the seven `python3` hook tests in `dbflux_core`, four process tests in `dbflux_lua`, two data-dir tests in `dbflux_ipc`/`dbflux_ssh`, and the six #625 document-cleanup tests above.
- `cargo nextest run -p dbflux_ui_document -E "test(save_as) or test(save_for_close) or test(landed_save) or test(failed_save)"` — 13 passed.

The cancellation test verifies that an untitled dirty buffer stays open and dirty, the close intent is dropped, and a later successful save does not close the tab. The chosen-path test verifies that the file receives the captured buffer, that the document retargets, and that the close the dialog interrupted is requested; the toolbar test verifies the same without a close request.

## Where was this tested?

- [x] Automated tests
- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] X11
- [ ] Wayland
- [ ] Other:

## Checklist

No CHANGELOG entry: this is an internal test seam/refactor with no user-visible behavior change.

## Labels to apply

`ui:feature`

Size note: the diff is ~553 changed lines, so per `CONTRIBUTING.md` this may need `size:exception` or splitting. The functional change is small; most of the diff is moving the duplicated save-dialog branches into one resolver and adapting call sites.


0xErwin1:
PR #625 landed and rewrote `save_file_as`, so this needs a rebase on `main`.

One thing the conflict markers do not show: your two new tests, `cancelling_save_as_keeps_the_tab_open` and `choosing_a_path_in_save_as_writes_and_closes`, now duplicate `an_untitled_dirty_document_saves_through_save_as_and_keeps_its_buffer_when_dismissed` and `save_as_writes_the_captured_content_and_retargets_the_document`, which are already in `main`.

PR #625 could not drive the picker, so its tests drop `_pending_save` for cancellation and call `enqueue_save_as_write` directly for Save As. Your seam is the better mechanism; I would keep one set of tests through it and drop the other.

Rebase and ping me.

xilec:
Rebased onto `main` (now includes #625 and #630) and consolidated the tests, as `294d517`.

**The rebase kept your tail.** `save_file_as` now resolves the target through `resolve_save_target` and hands the write to `enqueue_save_as_write`, so it gets #625's write queue, retargeting, scratch cleanup and fallback toast — my older inline write path was dropped by the merge rather than kept alongside.

**Test consolidation**, all through the `SaveTargetProvider` seam:

- `an_untitled_dirty_document_saves_through_save_as_and_keeps_its_buffer_when_dismissed` is gone. `cancelling_save_as_keeps_the_tab_open` reaches the same outcome through a real `SaveTargetOutcome::Cancelled` and covers more: the buffer stays dirty, the close intent is dropped, no file is created, and a later save lands without asking to close. The `Deferred` disposition the dropped test also asserted is still covered by `a_second_untitled_close_while_save_as_is_pending_starts_no_second_flow`.
- `save_as_writes_the_captured_content_and_retargets_the_document` is gone. Its one claim the seam-driven test did not make — the document is retargeted at the chosen path — is now asserted by `choosing_a_path_in_save_as_writes_and_closes`.
- `save_as_from_the_toolbar_retargets_without_asking_to_close` is new, so the direct entry point stays covered through the seam: driving `save_file_as` (not the close flow) writes the captured buffer, retargets the document, and emits no `RequestClose`.

Validation: `cargo fmt --all` and `cargo clippy --workspace -- -D warnings` clean; `cargo nextest run -p dbflux_ui_base -p dbflux_ui_document -p dbflux_ui_windows -p dbflux_ui` → 2031 tests, 2025 passed, 3 skipped, 6 failed (the `dbflux_ui` document-cleanup tests that seed a file under this sandbox's read-only `$HOME`, which came in with #625); `cargo nextest run --workspace --no-fail-fast` → 6459 passed, 249 skipped, 19 failed, all of them environment-dependent here.

The PR body has been updated with the above.

0xErwin1 (approved):
Approved. Verified against `main` at `560a8933`.

I applied this diff and ran `cargo nextest run -p dbflux_ui_base -p dbflux_ui_document -p dbflux_ui_windows -p dbflux_ui`: 2031 tests, 2030 passed, 1 failed, 3 skipped. The only failure is `palette_filtering_large_dataset_completes_within_budget`, which also fails on `main` alone. So the 6 cleanup failures in the validation section are specific to your sandbox rather than an expected outcome of this change, and the section is worth correcting before it is read as a regression.

One thing that changed on the way in: the `rfd::FileDialog` to `rfd::AsyncFileDialog` switch in `connection_manager/export_modal.rs` is a behaviour change, not a refactor detail, so it is recorded in the squash commit message instead of staying invisible under the title.

The rebase and the test consolidation both look right. Nothing lost from the two dropped tests: the retarget assertion moved into the seam-driven test and the `Deferred` assertion survives in the second-close test.

PR #628, opened 2026-09-17
fix(core): bound keyring calls so a silent secret service cannot freeze the UI

xilec:
## Summary

A keyring call is now bounded: `KeyringSecretStore` runs every operation —
availability probe, `get`, `set`, `delete` — on a worker thread and gives up
after five seconds with `DbError::Timeout` instead of blocking the caller
forever. The primitive (`dbflux_core::core::blocking::run_with_timeout`) sits
next to the crate's other small cross-cutting helpers; it is `pub(crate)`, so it
serves any blocking call inside `dbflux_core` that has neither an async API nor a
timeout of its own. A write that let the bound expire puts further writes on a
five-second hold, so a burst of clicks on Save cannot each pay the bound and
leave another abandoned worker behind, and a probe that times out leaves the
store available instead of disabling secrets for the rest of the process.

## What does this resolve?

- No linked issue: the freeze was reproduced by hand on a niri/Wayland desktop
  (log and bus trace below) rather than filed first — happy to open one if you
  want it tracked.

On the connection-manager save path a keyring write runs on the GPUI main thread
(`ConnectionManagerWindow::save_profile` → `AppStateEntity::update` →
`save_password`), and the `keyring` crate is synchronous with no timeout. So a
secret service that accepts a request and never answers freezes the whole
application:

* the window stops repainting (the event loop is inside the blocking call),
* the IPC socket stops answering, and
* the profile is never committed, because the password write happens before the
  profile is stored.

Reproduced on a niri/Wayland desktop. A debug run stopped at:

```
07:18:00  Saving profile: testdb, save_password=true, password_len=6
07:18:00  Saving password to keyring for profile 92833710-…
07:18:00  keyring: set password for entry SsCredential { … }   <- nothing after this
```

with `cfg_connection_profiles` empty, no audit event after startup, every thread
idle or waiting (no `D` state) and CPU nearly flat — an unresponsive write, not a
busy loop.

## How was this solved?

`KeyringSecretStore` wraps every call it makes in
`dbflux_core::core::blocking::run_with_timeout`: the call runs on a
`dbflux-blocking` worker and the caller waits on a channel with `recv_timeout`,
turning a service that never answers into `DbError::Timeout` after five seconds.
That error flows into the failure path already added in "Password save
failures are reported", so the form stays open and the user gets the error
instead of a frozen window. Calls that answer are unaffected — the bound is only
ever reached when nothing answers.

A worker that overruns the bound is abandoned rather than joined: Rust has no
safe way to kill a thread, and a call that cannot be cancelled is exactly what
this guards against. That is the tradeoff for a responsive application, and the
reason for the second half of the change: because the abandoned thread stays for
the life of the process, a store that has let one *write* time out stops writing
for the next five seconds. A burst of clicks on Save during the hold fails
immediately instead of paying the bound again and leaving another worker (and
another secret-service connection) behind; the hold then expires on its own, so
saving again after unlocking the keyring works without restarting the
application. Reads are deliberately not covered by the hold: in this failure
mode they answer in about a millisecond, and the session stays useful only while
stored passwords can still be read.

Why not a shared executor: `dbflux_core` has no async runtime (tokio is pulled in
with the `time` feature only, and `spawn_blocking` needs `rt`), and GPUI's
`background_executor` belongs to the UI layer, which `core` deliberately does not
depend on. A pooled thread would also be the wrong home for a hung call: it would
occupy a pool slot indefinitely, while a detached thread stays isolated.

Known limitations, deliberately out of scope: the profile/secret commit policy is
unchanged, so a failed write still leaves the profile uncommitted and the form
open, as the tests added with the previous change require; the save path is still
synchronous, so the user pays the five-second bound once per failed attempt
(making it fully asynchronous — a background executor plus a pending state in the
form — is a larger change); and why a locked collection blocks a write belongs to
the desktop, where unlocking the keyring (PAM at login, autostart) is the
user-side fix.

## Validation

In the running application, saving a connection with **Save password** checked
now costs the bound and reports it instead of freezing:

```
08:35:50.813  Saving password to keyring for profile a0f6a2ab-…
08:35:55.813  WARN Blocking call 'set' did not answer within 5s; giving up on it
08:35:55.813  WARN Secret service did not answer 'set'; DBFlux stops writing to the keyring …
08:35:57.471  Saving profile: testdb, save_password=true
08:35:57.695  Saving profile: testdb, save_password=true      <- clicks land ~200ms apart
08:35:57.891  Saving profile: testdb, save_password=true         and each fails at once,
08:35:58.024  Saving profile: testdb, save_password=true         never paying the bound
```

The window keeps repainting, the error toast appears, the form stays open, and
the window stays usable while the abandoned call is still outstanding.

Which call stops, captured on the session bus during the same Save:

```
OpenSession, SearchItems, ReadAlias -> /collection/login, SearchItems   <- reads answer
Unlock(/collection/login) -> prompt /org/freedesktop/secrets/prompt/u81
Prompt.Prompt(u81)                                                      <- never completes
```

The login collection was locked, and unlocking it asks the desktop for an
interactive prompt: the call then sits in `Prompt.Prompt` and never returns,
because that prompt is not answered. Reads stay fast because a locked collection
can still be searched, which is why the availability probe finds the service
present while every write blocks.

Live check of the bound on that desktop (temporary test, not committed):

```
available = true
set #1             -> Err(Timeout) after 5.000147457s
set #2             -> Err(Timeout) after 4.88µs
get                -> Ok(None)     after 1.676299ms
waiting out the 5s write hold...
set #3 (hold over) -> Err(Timeout) after 5.000133735s
```

Once the keyring was unlocked out of band, the same Save stored the password and
closed the form, and the same store answers in milliseconds — the bound is three
orders of magnitude away from a healthy keyring, so a service that answers never
sees it:

```
set       -> Ok("ok") after 6.94ms
get       -> Ok(…)    after 0.88ms
get-again -> Ok(…)    after 0.92ms
delete    -> Ok("ok") after 2.98ms
```

Unit tests: the bound returns a call's result, propagates its error, and gives up
on a hung call; the store routes its calls through its own bound; a silent
service is not asked to write twice while the hold lasts; writes resume after the
hold expires and a successful write clears it; a read is still attempted after a
write timed out; and a probe that never answers leaves the store available, while
a probe that reports a real failure still disables secrets.

A probe that times out is deliberately *not* treated as a missing backend. The
locked-keyring case is already handled this way for `NoStorageAccess`, and
latching `available = false` would drop the secrets already stored in the keyring
(SSH tunnels, proxies, auth-profile fields) for the rest of the process — `get`
would answer `Ok(None)` and writes would be skipped — while a timed-out write
heals by itself once the cooldown passes. With the store left available, reads
are still attempted and every call reports its own timeout.

`cargo fmt --all` and `cargo clippy --workspace -- -D warnings` are clean.
`cargo nextest run --workspace`: 6412 tests run — 6399 passed, 249 skipped, 13
failed. All 13 failures are pre-existing in this sandbox (no `python3` for the
hook and Lua process tests, read-only `~/.local/share` for `dbflux_ipc` and
`dbflux_ssh` path tests) and unrelated to the change.

## Where was this tested?

- [x] Automated tests
- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] X11
- [x] Wayland
- [x] Other: niri/Wayland session with `gnome-keyring` on `org.freedesktop.secrets`

## Checklist

## Labels to apply

`storage:bug`, `platform:linux`


0xErwin1:
One change before this merges.

On a probe timeout `check_availability` returns false, so `available` stays false for the rest of the process. That contradicts the comment above it, where `NoStorageAccess` is deliberately kept available so secrets are not silently dropped: present-but-not-answering is the locked case, not the `PlatformFailure` case. A hung write self-heals after the cooldown; a hung probe costs secrets until restart, dropping SSH/proxy secrets on a log line and returning `Ok(None)` from `get`. Leave availability unknown on a timeout and re-probe on the next write, or report available as `NoStorageAccess` does. I did not reproduce this branch, since in the mode you documented reads answer in about a millisecond.

Two nits, not blocking: `run_with_timeout` is `pub(crate)` inside a private `mod core`, so the body's "any blocking call can use it" holds only inside `dbflux_core`; and nothing tests the probe-timeout path.

Verified on 641924cd: `cargo nextest run -p dbflux_core` 1237/1237, clippy matches CI.

xilec:
Both points addressed in `1787296`, kept as a separate commit so the delta is easy to read.

**Probe timeout no longer disables secrets.** You're right that the branch contradicted the `NoStorageAccess` comment: it returned `false`, so `available` stayed false for the process, `get` answered `Ok(None)` and writes were skipped. It now reports available for the same reason `NoStorageAccess` does — a service that accepts the probe and never answers is the locked case, not a missing backend — with its own log line, and every call still reports its own timeout.

That also covers the "leave availability unknown and re-probe" option without a separate unknown state: the store stays available, so operations are attempted rather than skipped, and the next successful call is the proof the service is back. If you would rather have an explicit re-probe on the next write, I can shape it that way instead.

`check_availability` is split into `check_availability_with(probe)` so the branch is testable, and `a_probe_that_times_out_keeps_the_store_available` covers both sides: a silent probe stays available, a real failure still disables secrets.

`cargo nextest run -p dbflux_core` here: 1238 tests, 1231 passed, 7 failed — all seven are the `python3` hook tests this sandbox cannot run, and your 1237/1237 run is the same suite before the added test. `cargo clippy --workspace -- -D warnings` is clean. The PR body has been updated with the new numbers and the probe paragraph.

**Nits.** Fixed both: the PR body no longer claims the helper is callable outside `dbflux_core` (it is `pub(crate)`), and the probe-timeout path now has the test above.

0xErwin1 (approved):
Verified on `1787296f`: `cargo nextest run -p dbflux_core` is 1238/1238 here, so the seven `python3` hook failures are your sandbox as you said, clippy matches CI, and all 13 checks are green. Reporting the store available on a probe timeout is the side I would pick too, and `check_availability_with` is a clean seam for making it testable. Merging.

PR #636, opened 2026-09-17
fix(ui): keep a reopened table's grid editable

### LINKED ISSUE #634: [bug] Reopening a table in the same connection leaves the grid read-only
### DBFlux version

0.8.0-nightly+18eec1b

### Operating system

Linux (Wayland)

### OS / distro detail

NixOS 26.05, kernel 6.18; niri 26.04 compositor, native Wayland client (not XWayland). The behaviour is UI-side and does not depend on the compositor.

### Architecture

amd64 / x86_64

### Affected driver (if any)

postgres — observed on a PostgreSQL table. The mechanism is in shared UI code (a cache-key mismatch and a sentinel error path), so any driver whose table source carries a database name is affected.

### Steps to reproduce

1. Connect to a PostgreSQL profile. Do **not** click the database node in the sidebar: the current database's node is already expanded, and clicking it is one of the few actions that records an active database.
2. Open a table from the sidebar. The grid is editable — Enter starts the inline editor, the row/apply actions are available.
3. Close the tab and open the same table again. The grid is read-only: no editor starts, and the "no primary key" banner is shown.
4. Disconnect and reconnect the profile, then open the table once more — it is editable again, and closing and reopening it repeats the read-only state.

### Expected behavior

Reopening the same table in the same connection keeps the same editing capability. The table's primary key does not change between the first and the second open, and it is still present in the connection's cache.

### Actual behavior

Only the first open after a (re)connect is editable; every later open of the same table in that connection is read-only until the profile is reconnected.

**Cause.** The panel reads and writes `ConnectedProfile::table_details` under *different* database keys, and the write path treats "already cached" as an error.

Reading uses `active_database`:

```rust
// crates/dbflux_ui_document/src/data_grid_panel/mod.rs:2580
let database = connected.active_database.as_deref().unwrap_or("default");
let cache_key = (database.to_string(), table.schema.clone(), table.name.clone());
if let Some(table_info) = connected.table_details.get(&cache_key) { /* … */ }
```

The fetch uses the source's database instead:

```rust
// crates/dbflux_ui_document/src/data_grid_panel/mod.rs:713-733
let database = source_database.unwrap_or_else(|| {
    /* … active_database … */ .unwrap_or_else(|| "default".to_string())
});
let params = match self.app_state.read(cx).prepare_fetch_table_details(
    profile_id, &database, table.schema.as_deref(), &table.name,
) { … };
```

`prepare_fetch_table_details` returns a sentinel error when the entry is already there, and it keys it the same way the fetch does:

```rust
// crates/dbflux_core/src/connection/manager.rs:1500-1508
let cache_key = (database.to_string(), schema.map(String::from), table.to_string());
if let Some(details) = connected.table_details.get(&cache_key)
    && (details.columns.is_some() || details.sample_fields.is_some())
{
    return Err("Table details already cached".to_string());
}
```

and the panel abandons the fetch on any error:

```rust
// crates/dbflux_ui_document/src/data_grid_panel/mod.rs:736-746
let params = match … {
    Ok(p) => p,
    Err(e) => {
        log::warn!("[PK] Failed to prepare fetch_table_details: {}", e);
        return;
    }
};
self.pk_details_pending = true;
```

For PostgreSQL the two keys can never meet: `active_database` is `None`, while the table's source database is `Some("testdb")`.

- Connecting leaves it unset — `crates/dbflux_core/src/connection/manager.rs:791-799` (`active_database: None`). It is set only by sidebar click handlers (`crates/dbflux_ui_sidebar/src/operations/tree_ops.rs:802, 974, 1079`).
- The current database's node is expanded without a click (`crates/dbflux_ui_sidebar/src/tree_builder.rs:1926-1930`), and its tables are built carrying `target_database = Some(db_name)` (`tree_builder.rs:1759-1771`). So a table can be opened straight from the tree while no active database was ever recorded.

The cycle follows:

1. **First open.** The reader looks under `("default", …)`, misses, so `fetch_table_details_for_pk` runs, `prepare_fetch_table_details` finds nothing, the fetch writes the entry under `("testdb", …)` and the panel sets `pk_columns` from it (`mod.rs:787`, `:808-810`) — the grid is editable.
2. **Second open.** The reader looks under `("default", …)` again and misses again, but now `prepare_fetch_table_details` finds the entry that the first open wrote and returns `"Table details already cached"`. The panel logs a warning and returns **before** touching `pk_details_pending` or `pk_columns`.
3. With `pk_columns` empty, `mutations_enabled()` is `false` (`mod.rs:3052-3057`, gated on `!self.pk_columns.is_empty()`), `pk_indices` is left empty (`mod.rs:2218-2228`), `DataTableState::set_pk_columns` clears `is_editable` (`crates/dbflux_components/src/components/data_table/state.rs:635-637`), and `start_editing` refuses the gesture (`state.rs:727-740`).
4. Reconnecting clears `table_details` and resets `active_database`, which is why editability comes back and the cycle repeats.

The primary key is in the cache the whole time; only the key used to look it up differs. The codebase already treats the same sentinel as "the details are here, read them" elsewhere — `crates/dbflux_ui_document/src/migrate_wizard/mod.rs:175-199` reads the cache on that error — while the grid panel treats it as a failure.

Not `EditableBinding`: that path only gates builder-generated SELECT results (`crates/dbflux_core/src/query/visual_query.rs:580-593`, `data_grid_panel/query.rs:420-430`). A plain `DataSource::Table` grid is gated on `pk_columns` alone, and `MutationPolicy`/`read_only_reason` are only consulted when a change is applied (`mod.rs:4077-4086`), not for starting an edit.

To be explicit about what is observed and what is read from the code: the read-only second open and the recovery on reconnect were seen in the app; the state that triggers it (`active_database` unset, and no click on the database node) is what the code above requires, and is inferred from it rather than captured from that session.

### Logs / stack traces

No log was captured from the session. The failing path logs

```
[PK] Failed to prepare fetch_table_details: Table details already cached
```

at `crates/dbflux_ui_document/src/data_grid_panel/mod.rs:741`, so that line in the application log confirms this mechanism.

### Labels I plan to apply

- [x] Kind label — `ui:bug`
- [x] Driver label if applicable — `driver:postgres` (where it was observed; the mechanism itself is not driver-specific)
- [ ] Platform label if platform-specific (`platform:linux`, …) — not applicable
- [x] Data-model label if relevant — `kind:sql`

I am not a maintainer of this repository and cannot apply labels from a fork, so the boxes above state what the issue needs rather than what I can set: please add `ui:bug`, `driver:postgres` and `kind:sql` during triage.

xilec:
## Summary

A table reopened in the same connection keeps its grid editable. The panel's readers and the fetch that fills the connection's table-details cache now build the cache key the same way, and details that are already cached are used instead of being fetched again.

## What does this resolve?

- Resolves #634

## How was this solved?

The panel wrote table details into the connection cache under the table source's database but read them under `ConnectedProfile::active_database`. Those differ in a common case: connecting leaves `active_database` unset, and a table opened straight from the sidebar tree carries its own database — clicking the database node is what records an active database. On the first open the lookup missed the cache, the fetch wrote the entry under the table's database, and the grid was editable. On the second open the lookup missed it again while `prepare_fetch_table_details` recognised its own entry and returned the `"Table details already cached"` sentinel, which the panel treated as a failure and returned from before setting `pk_columns`. With no primary key, `mutations_enabled()` is false and the grid stays read-only until the profile is reconnected.

- `DataGridPanel::table_details_database` builds the key from the table's own database, falling back to the connection's active database and then `"default"` — the same expression the fetch writes the entry under, and the one `compute_builder_binding` already used. Every reader of the cache in this module now goes through it: the primary-key/`ORDER BY` lookup, column defaults, cached column details, FK hints, the FK reference list in the row inspector, the row SQL preview's PK indices, and the builder binding. Each of them missed the cache in the same scenario, so they are fixed together rather than left to disagree with the primary-key path.
- `fetch_table_details_for_pk` checks for cached columns under that key before preparing a fetch and applies them through the extracted `apply_pk_details`, so a warm cache never reaches the sentinel error path at all.
- `primary_key_names` replaces four copies of the same column filter.
- `reopened_table_finds_cached_primary_key_without_an_active_database` covers the lookup and the fetch path. Most of the diff is that test plus the `StubConnection` stub it shares with the existing `cold_cache_upgrades_to_editable_on_table_details_arrival` test, which previously carried its own copy of it; the production change is roughly 60 lines plus mechanical call-site updates.

No driver-specific code is involved — the change is in shared UI code.

Known limitation, not introduced here: the sidebar writes the same cache through `ItemIdParts::cache_database`, which falls back to the *schema* name when a table node carries no database (`crates/dbflux_ui_sidebar/src/lib.rs:745`), while this helper falls back to the active database. For a collapsed/single-database node with no active database the two keys can still differ, and the panel re-fetches instead of hitting the cache — a wasted fetch, not a read-only grid. Making the key a single shared function for both crates would close it; that is a follow-up, not part of this fix.

## Validation

- `cargo fmt --all` — clean.
- `cargo nextest run -p dbflux_ui_document` — 1324 passed, 3 skipped.
- Manual check on a dev build (`0.8.0-dev.0`, PostgreSQL on Wayland): with the profile reconnected so no active database is recorded, a table opened from the tree was editable, and closing and reopening the tab left it editable — the read-only second open from the report no longer reproduces, and the `Table details already cached` warning no longer appears in the application log.
- The regression test fails against the old lookup with the exact symptom: `left: []`, `right: ["id"]` at the "cached details must be found under the table's database, not the active one" assertion, and `mutations_enabled()` is false in that state.

`CHANGELOG.md` is deliberately untouched: CONTRIBUTING.md § "Commit messages are load-bearing" says not to hand-edit it or `[Unreleased]`, because git-cliff generates both from commit messages. The commit subject is `fix(ui): keep a reopened table's grid editable`, which is the changelog bullet.

## Where was this tested?

- [x] Automated tests
- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] X11
- [x] Wayland
- [ ] Other:

## Checklist

## Labels to apply

- Kind: `ui:bug`
- Data model: `kind:sql`

I am not a maintainer of this repository and cannot apply labels from a fork, so these state what the PR needs rather than what I can set: please add them during triage.


crates/dbflux_ui_document/src/data_grid_panel/utils.rs
```diff
@@ -44,11 +44,32 @@ pub(super) fn value_to_json(value: &Value) -> serde_json::Value {
...
 impl DataGridPanel {
+    /// Database the connection's `table_details` cache is keyed by for the table
+    /// this panel shows.
+    ///
+    /// The table's own database leads — a table opened from the sidebar carries
+    /// it even when the connection recorded no active database, because
+    /// connecting leaves `active_database` unset and only clicking the database
+    /// node sets it. This is the key the fetch that fills the cache writes with,
+    /// so every reader must build it the same way.
+    pub(super) fn table_details_database(
```
0xErwin1, line 55: Good shape, and the doc comment finally records why the table's own database leads. Two follow-ups in the same direction: this publishes the key component only, so the five call sites still repeat the destructure, the key tuple and the lookup, and a sixth reader can drift the same way this one did. A `fn table_details_for(&self, cx) -> Option<&TableInfo>` on top of it would make the wrong key impossible to write. The `String` return allocates on every lookup, which is fine at these call sites (per rebuild, not per cell); worth knowing only so nobody lifts it into a render path.

crates/dbflux_ui_document/src/data_grid_panel/navigation.rs
```diff
@@ -94,8 +94,13 @@ impl DataGridPanel {
                 total_rows,
                 ..
             } => {
-                let pk_order =
-                    Self::get_primary_key_columns(&self.app_state, *profile_id, table, cx);
+                let pk_order = Self::get_primary_key_columns(
+                    &self.app_state,
+                    *profile_id,
+                    database.as_deref(),
```
0xErwin1, line 100: Right. Worth knowing that this shape is the model for the fix I would like in `apply_pk_details`: recompute `pk_order`, rewrite `self.source` with it, and queue a `PendingRequery`.

crates/dbflux_ui_document/src/data_grid_panel/context_menu/mod.rs
```diff
@@ -3124,9 +3126,12 @@ impl DataGridPanel {
             None => return,
         };
 
-        let database = connected.active_database.as_deref().unwrap_or("default");
+        let source_database = match &self.source {
```
0xErwin1: This second match over `self.source` is unnecessary, since the function already destructured it at 3106-3111. Extending that one destructure to `(profile_id, table_ref, database)` keeps every reader in this PR looking the same, which is the point of the change.

crates/dbflux_ui_document/src/data_grid_panel/mod.rs
```diff
@@ -715,21 +716,38 @@ impl DataGridPanel {
...
 
-        let database = source_database.unwrap_or_else(|| {
-            let state = self.app_state.read(cx);
-            state
-                .connections()
-                .get(&profile_id)
-                .and_then(|c| c.active_database.clone())
-                .unwrap_or_else(|| "default".to_string())
-        });
+        let database = match self.app_state.read(cx).connections().get(&profile_id) {
```
0xErwin1: The `None` arm invents `"default"` for a profile that is not connected, but `prepare_fetch_table_details` rejects that case anyway with `Profile not connected` (`manager.rs:1497`). Returning early would keep the real reason visible instead of dressing it up as a database name.

crates/dbflux_ui_document/src/data_grid_panel/mod.rs
```diff
@@ -715,21 +716,38 @@ impl DataGridPanel {
...
+            None => source_database.unwrap_or_else(|| "default".to_string()),
+        };
 
         log::info!(
             "[PK] Fetching table details for PK columns: {}.{}",
             database,
             table.qualified_name()
         );
 
+        // Details cached under this key must be used rather than fetched again:
```
0xErwin1, line 733: The `log::info!` above still fires before this check, so a warm cache logs `Fetching table details for PK columns` immediately followed by `Using cached table details`. That is the line #634 used as diagnostic evidence, so move it below this block.

crates/dbflux_ui_document/src/data_grid_panel/mod.rs
```diff
@@ -715,21 +716,38 @@ impl DataGridPanel {
...
+        // `prepare_fetch_table_details` refuses a second fetch for a key that is
+        // already populated, and treating that refusal as a failure is what left
+        // a reopened table read-only.
+        let cached_pk_names = self
```
0xErwin1, line 739: This check only covers `columns.is_some()`, so an entry carrying just `sample_fields` still falls through to the sentinel and the silent `return` below. `migrate_wizard/mod.rs:153-199` already treats that sentinel as "read the cache"; doing the same here closes the shape you left open and makes the comment above true without qualification.

crates/dbflux_ui_document/src/data_grid_panel/mod.rs
```diff
@@ -715,21 +716,38 @@ impl DataGridPanel {
...
+            .app_state
+            .read(cx)
+            .get_table_details(profile_id, &database, table.schema.as_deref(), &table.name)
+            .and_then(|details| details.columns.as_deref())
+            .map(Self::primary_key_names);
+
+        if let Some(pk_names) = cached_pk_names {
```
0xErwin1, line 745: This `return` is the right fix, but the `Err` arm below it still returns silently, which is how #634 presented to the user: a read-only grid and nothing but a log line. Now that the sentinel can only arrive from the `sample_fields` shape, anything else in that arm, including a disconnected profile, should go through `report_error`, so the grid never loses editability without a signal.

crates/dbflux_ui_document/src/data_grid_panel/mod.rs
```diff
@@ -801,54 +814,61 @@ impl DataGridPanel {
...
+                    self.compute_builder_binding(spec.as_ref(), pid, database.as_deref(), cx);
+                self.pk_columns = binding
+                    .as_ref()
+                    .map(|b| b.pk_columns.clone())
+                    .unwrap_or_else(|| self.pk_columns.clone());
+                self.builder.builder_editable_binding = binding;
+            }
+        }
+
+        self.pending.rebuild = true;
```
0xErwin1: This sets `pending.rebuild`, and `rebuild_table` only re-reads PK indices (`render.rs:217`, `mod.rs:2312`), so the panel never gets an `ORDER BY` it did not already have. On the first open after connect the cache is cold, `get_primary_key_columns` returns empty at 654, and every page of that panel is issued as `LIMIT/OFFSET` with no ordering (`dbflux_core/src/query/table_browser.rs:419`); MSSQL is the only driver that injects a fallback (`driver.rs:2211`). Rows can repeat or be skipped while paging a large table. `navigation.rs:94-131` is the pattern to copy: when `order_by` is empty, rewrite the source with the PK order and queue a `PendingRequery`. That makes the first open as deterministic as the reopen this PR fixes. While here, 848-856 opens two matches over the same source and 860-863 self-clones; one destructure and an `if let Some(b) = &binding` is enough.

crates/dbflux_ui_document/src/data_grid_panel/mod.rs
```diff
@@ -2566,9 +2586,16 @@ impl DataGridPanel {
 
     // === Helpers ===
 
+    /// Primary-key columns for `table`, read from the connection's cached table
+    /// details.
+    ///
+    /// `database` is the database the table was opened from. The cache key is
+    /// built by [`DataGridPanel::table_details_database`], the same key
```
0xErwin1, line 2678: `fetch_table_details_for_pk` is the key this reader has to agree with, and now it does. The sidebar builds a different one: `ItemIdParts::cache_database` (`dbflux_ui_sidebar/src/lib.rs:745`) falls back to the schema name when a node carries no database, which is the `build_schema_children(..., None, ...)` path at `tree_builder.rs:358-373`. There the panel uses `("default", ...)` and the sidebar `("main", ...)`, so one table ends up with two entries that can go stale independently. Not a read-only path, since SQLite ignores the database argument (`driver.rs:878-881`) and the fetch succeeds, and it matches the limitation you flagged. What makes it unfixable by one fallback is that the string carries two roles at once, cache key component and fetch target database. Worth saying that here, and worth noting that `new_internal` at 898-903 is the one reader in this module still on a third chain (through `table.schema`).

0xErwin1 (commented):
The key alignment is correct and the warm-cache check is what actually defuses the sentinel. `apply_pk_details` is order-equivalent to the block it replaces, and the test asserts the exact state the old lookup produced. I did not build or run the branch; this review is static against `538bd15` and the diff.

Nothing here blocks on correctness. Two small changes I would ask for before merge: the `Err` arm in `fetch_table_details_for_pk` still returns silently, which is how #634 reaches the user, and the `log::info!` above the new cache check now prints `Fetching` on a cache hit. The `ORDER BY` the first open never gets is the same class of bug in the same function, so I would take it in this PR. Everything else is inline.

`CHANGELOG.md` untouched is right, per CONTRIBUTING.md on commit messages driving the generated changelog, and `ui:bug` plus `kind:sql` are the right labels. `driver:postgres` should stay off this PR since no driver code changes.

crates/dbflux_ui_document/src/data_grid_panel/mod.rs
```diff
@@ -4756,6 +4763,81 @@ mod tests {
...
+
+        fn kind(&self) -> dbflux_core::DbKind {
+            dbflux_core::DbKind::SQLite
+        }
+
+        fn schema_loading_strategy(&self) -> dbflux_core::SchemaLoadingStrategy {
+            dbflux_core::SchemaLoadingStrategy::SingleDatabase
+        }
+
+        fn dialect(&self) -> &dyn dbflux_core::SqlDialect {
```
0xErwin1, line 4925: `unimplemented!()` here means any path under test that needs the dialect panics instead of failing. Say so in the stub's doc comment so the next test knows the limit.

crates/dbflux_ui_document/src/data_grid_panel/mod.rs
```diff
@@ -6895,6 +6910,181 @@ mod tests {
...
+            *found.borrow(),
+            vec!["id".to_string()],
+            "cached details must be found under the table's database, not the active one"
+        );
+
+        let panel = panel_holder
+            .borrow()
+            .clone()
+            .expect("panel should be created");
+        let edits_allowed = window.update(|_, app| panel.read(app).mutations_enabled());
```
0xErwin1, line 7178: `mutations_enabled()` is the panel gate; the grid gesture #634 is about reads `DataTableState::is_editable`, which `rebuild_table` sets from these PKs (`mod.rs:2312`). The existing tests call `panel.rebuild_table(None, cx)` and then assert `table_state.read(app).is_editable()` (`mod.rs:7339-7355`); two more lines here would pin the reported symptom. The `order_by` computed at 7006 is also never asserted, and that is what would catch the missing ORDER BY on the cold path.

crates/dbflux_ui_document/src/data_grid_panel/utils.rs
xilec: Taken in `cb4f240`: `DataGridPanel::table_details_for` is the accessor, and the five call sites plus `handle_generate_sql` go through it, so a sixth reader cannot pick a different key. The `String` stays as it is — per rebuild, not per cell — but the note is useful for anyone who lifts it into a render path.

crates/dbflux_ui_document/src/data_grid_panel/navigation.rs
xilec: Followed exactly. `apply_pk_details` now recomputes the key order, rewrites `self.source` with it when the panel has no order of its own, and queues `PendingRequery` (`58903b4`). `first_open_requeries_with_the_primary_key_order_it_learned` asserts the rewritten source and the queued requery, and fails without the change.

crates/dbflux_ui_document/src/data_grid_panel/context_menu/mod.rs
xilec: Gone in `cb4f240`. `handle_generate_sql` reads the database through `table_details_for`, so the second `match self.source` is no longer there to destructure.

crates/dbflux_ui_document/src/data_grid_panel/mod.rs
xilec: Fixed in `00699e4`. The `None` arm logs `[PK] Cannot fetch table details: profile … is not connected` and returns, so the real reason stays visible instead of being dressed up as a database name.

crates/dbflux_ui_document/src/data_grid_panel/mod.rs
xilec: Moved below the cache check in `35477e6`, with a comment recording that the line was read as evidence in #634 and why its old placement was misleading.

crates/dbflux_ui_document/src/data_grid_panel/mod.rs
xilec: Widened in `35477e6`: any entry under the key counts, so a `sample_fields`-only entry is a cache hit and the sentinel cannot arrive from that shape any more.

crates/dbflux_ui_document/src/data_grid_panel/mod.rs
xilec: Now reported through `report_error(UserFacingError::new(ErrorKind::Driver, labels::pk_details_fetch_failed_error(..)))` (`00699e4`), so the grid cannot lose editability without a signal. The disconnected profile is handled in the `None` arm above it, which logs and returns — there is no database to fetch from.

crates/dbflux_ui_document/src/data_grid_panel/mod.rs
xilec: Taken in this PR (`58903b4`). With `order_by` empty the source is rewritten with the PK order and `PendingRequery` is queued; a sort the user already chose is left alone, and a table whose keys are still unknown rebuilds as before. `first_open_requeries_with_the_primary_key_order_it_learned` pins both and fails against the old `pending.rebuild`. The two matches are one destructure now (`72d658e`): `apply_pk_details` reads `self.source` once for both the binding recompute and the requery decision, and the recomputed binding is applied with `if let Some(binding) = &binding` rather than `unwrap_or_else(|| self.pk_columns.clone())`.

crates/dbflux_ui_document/src/data_grid_panel/mod.rs
xilec: Both recorded on `get_primary_key_columns` (`e653d9d`): the sidebar's `ItemIdParts::cache_database` fallback to the schema name (`main` against `default`), the fact that the string carries two roles at once — cache key component and fetch target database — which is what makes one shared fallback impossible, and `new_internal` as this module's third chain. Thanks for the pointer to `tree_builder.rs:358-373`; that is the path the note names.

crates/dbflux_ui_document/src/data_grid_panel/mod.rs
xilec: Documented on `struct StubConnection` (`063cf3b`), along with why the tests can drive `new_internal` and the method under test directly instead of a path that needs the dialect.

crates/dbflux_ui_document/src/data_grid_panel/mod.rs
xilec: Both in `da405b5`. The reopened-table test calls `rebuild_table` and asserts `table_state.read(app).is_editable()` next to `mutations_enabled()`, and asserts the `order_by` the cache produced (`["id"]`). The cold-path order is asserted in the new test, which is what fails without the requery.

0xErwin1 (approved):
Approved. Every review thread is addressed in the branch. I re-checked the two comments that carry the fix: the requery cannot loop and cannot drop staged edits, because the grid is still not editable when it fires, and the warm-cache read now goes through the single accessor every reader shares. CI is green on 72d658e9. Two residuals are recorded in the earlier comments and do not block: the cache-hit test is now wider than the populated guard in prepare_fetch_table_details, and the requery keeps the current page offset where handle_sort_clear resets it.

PR #637, opened 2026-09-17
fix(ui): let an inline edit be reverted to the row's own value

### LINKED ISSUE #633: [bug] Inline cell edit cannot be reverted to the value already in the row
### DBFlux version

0.8.0-nightly+18eec1b

### Operating system

Linux (Wayland)

### OS / distro detail

NixOS 26.05, kernel 6.18; niri 26.04 compositor, native Wayland client (not XWayland). The behaviour is UI-side and does not depend on the compositor.

### Architecture

amd64 / x86_64

### Affected driver (if any)

postgres — observed on a PostgreSQL table, but the cause is in the driver-agnostic grid component, so any mutable grid is affected.

### Steps to reproduce

1. Open a table in the data grid and double-click (or press Enter on) a text cell whose value is `alice`.
2. Replace it with `carol` and commit the edit (Enter).
3. Edit the same cell again, replace with `dave`, commit.
4. Edit the same cell a third time, type `alice` — the original value — and commit.

### Expected behavior

The cell shows `alice` again and the row is no longer dirty: reverting an edit to the value the row already has in the database is the same as having made no edit at all. The pending-change counter for that row goes back to zero, and no `UPDATE` is generated for it.

### Actual behavior

The cell keeps showing `dave` — the value entered in step 3. The row stays marked as changed, and applying the changes writes `dave`, so the value typed in step 4 is silently lost.

The same happens in the other direction after a longer chain: any attempt to go back to the stored value leaves whichever pending value was there before.

**Cause.** The inline text editor commits in `close_editor` (`crates/dbflux_components/src/components/data_table/state.rs:987-998`). It compares the typed string with the *base* value of the model and skips the write when they are equal:

```rust
Some(VisualRowSource::Base(base_idx)) => {
    let original = self.model.cell(base_idx, coord.col)
        .map(|c| c.display_text().to_string())
        .unwrap_or_default();

    if value_str != original {
        let cell_value = super::model::CellValue::text(&value_str);
        self.edit_buffer.set_cell(base_idx, coord.col, cell_value);
    }
}
```

Equality here means "this edit is a no-op", but it is treated as "there is nothing to do". The pending override from step 3 is neither replaced nor removed, and the grid renders from the override, not from the base value:

- `EditBuffer::get_cell` returns `self.overrides.get(&(row, col)).unwrap_or(base)` (`crates/dbflux_components/src/components/data_table/model.rs:501-503`), and the table draws through it (`table.rs:1094-1105`, `record.rs:249-256`).
- Applying the row sends `row_changes`, which is built from the override map (`crates/dbflux_ui_document/src/data_grid_panel/mutations.rs:375-380`), so `dave` is what reaches the database.
- The dirty marker and the pending-edit counter are driven by the same map (`RowState::Dirty`, set only in `EditBuffer::set_cell`, `model.rs:511-523`; counted by `dirty_row_count`, `model.rs:544-546`), so the row also stays visibly dirty.

There is no way to fix this from the current buffer API: `EditBuffer` has `set_cell`, `clear_row` and `clear_all` (`model.rs:511-536`) but no way to drop the override of a single cell, and `close_editor` has no branch that expresses "this cell is back to its stored value".

Reopening the editor pre-fills it with the pending value (`state.rs:742-746`), which is why the user types the stored value and sees a different one retained — the guard rejects the write and the pre-existing override wins.

Not specific to the inline editor: the modal value editor writes through `set_cell` unconditionally (`crates/dbflux_ui_document/src/data_grid_panel/context_menu/mod.rs:2174-2200`), so it does not exhibit this, while the inline editor does. No driver-specific code is involved — this is entirely in `dbflux_components`.

### Logs / stack traces

None — no error is reported, the edit just does not take effect.

### Labels I plan to apply

- [x] Kind label — `ui:bug`
- [ ] Driver label if applicable (`driver:postgres`, …) — not applicable: the cause is in the shared grid component, so every mutable grid is affected
- [ ] Platform label if platform-specific (`platform:linux`, …) — not applicable
- [x] Data-model label if relevant — `kind:sql` (observed on a relational table; the mechanism is not specific to SQL)

I am not a maintainer of this repository and cannot apply labels from a fork, so the boxes above state what the issue needs rather than what I can set: please add `ui:bug` and `kind:sql` during triage.

xilec:
## Summary

An inline cell edit can be taken back by typing the value the row already holds: the pending change is dropped, the cell stops being shown as edited, and the row stops counting as modified. The same holds for the enum dropdown, the cell editor panel, "Set NULL", "Set default" and paste.

## What does this resolve?

- Resolves #633

## How was this solved?

The commit path compared the typed text with the value the row holds and only staged it when the two differed:

```rust
if value_str != original {
    self.edit_buffer.set_cell(base_idx, coord.col, cell_value);
}
```

When a value was already staged for that cell, typing the row's own value matched `original`, so nothing happened — and the previously staged value stayed in place. The grid kept showing the edited value and the row stayed marked as modified, so the change could only be taken back through undo, which is not the same action. The comparison is right, but its other branch had to drop the staged value instead of doing nothing.

- `DataTableState::stage_base_cell_value` is now the one place a value typed or picked for an existing row is staged: it compares the incoming value's display text with the row's own value and either clears the staged value or stages the new one.
- `EditBuffer::clear_cell` removes the override, returns the row to clean when it has no other staged cell (leaving a pending delete alone), and records `EditAction::ClearCell { row, col, replaced_value }`, so undo brings the dropped value back and redo drops it again.
- Every path that stages a value for an existing row goes through that seam: the inline editor (`close_editor`), the enum dropdown (`apply_enum_selection_at`), the cell editor panel (`write_cell_value`), the context menu's "Set NULL" and "Set default", and paste. Each of them had the same symptom.

Paste also carried a second problem: it used the selection's visual row index as a source row index, while the edit buffer is keyed by source rows — the render path states the split explicitly ("Use visual_ix for selection/display, but data_row_ix for edit buffer access"). With a pending insert above the selection, or after a pending insert was added, the value went to another row. Resolving the row through the visual order, as the neighbouring staging paths already do, fixes that; pasting into a pending insert now stages on the insert itself instead of on a row of the table.

Paste is reachable from the context menu only: the grid has no `Ctrl+V` binding, and the `Paste` event declared in `crates/dbflux_components/src/components/data_table/events.rs` has no emitter. Adding a shortcut is a separate change and this PR does not add one.

Out of scope: `clear_row` and `clear_all` push no undo entry at all, so the toolbar's clear actions are not undoable on their own. That is a pre-existing gap, unrelated to taking back a typed value, and this change does not touch it.

## Validation

- `cargo fmt --all` — clean.
- `cargo nextest run -p dbflux_components -p dbflux_ui_document` — 1760 passed, 3 skipped.
- New tests: `clear_cell_drops_one_staged_value_and_is_undoable` (buffer: one cell cleared, siblings kept, row back to clean, undo restores the dropped value, redo drops it again, clearing a clean cell records no undo step) and `typing_the_rows_own_value_drops_the_staged_edit` (the real editor: `start_editing` → type → `stop_editing`, staged once, then typed back).
- Red proof: with the comparison replaced by an unconditional `set_cell`, `typing_the_rows_own_value_drops_the_staged_edit` fails on "typing the value the row already holds must drop the pending change"; the fix was restored afterwards.

`CHANGELOG.md` is deliberately untouched: CONTRIBUTING.md says not to hand-edit it or `[Unreleased]`, because git-cliff generates both from commit messages. The commit subject is `fix(ui): let an inline edit be reverted to the row's own value`, which is the changelog bullet.

## Where was this tested?

- [x] Automated tests
- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] X11
- [x] Wayland
- [ ] Other:

## Checklist

## Labels to apply

- Kind: `ui:bug`
- Data model: `kind:sql`

I am not a maintainer of this repository and cannot apply labels from a fork, so these state what the PR needs rather than what I can set: please add them during triage.


crates/dbflux_components/src/components/data_table/model.rs
```diff
@@ -894,6 +919,20 @@ impl EditBuffer {
                 });
             }
 
+            EditAction::ClearCell {
+                row,
+                col,
+                replaced_value,
+            } => {
+                self.overrides.insert((row, col), replaced_value.clone());
+                self.row_states.insert(row, RowState::Dirty);
```
0xErwin1: `RowState::Dirty` is written unconditionally here, and it overwrites a pending delete. `clear_cell` (545) and the `redo` arm (1038) both guard this with `is_pending_delete`, and the pre-existing `SetCell` undo arm leaves row state alone when it restores a value, so this is the only arm that can move a row out of `PendingDelete`.

The sequence is reachable from the grid: stage a cell in row R, use the row's context menu `Delete row` (`mark_for_delete` at `context_menu/mod.rs:2771` sets `PendingDelete` and keeps the overrides), revert that cell to the value the row already holds so `clear_cell` runs and deliberately keeps `PendingDelete`, then Ctrl+Z. The row lands in `dirty_rows()` instead of `pending_delete_rows()`, and `request_save_all` (`state.rs:1100-1114`) emits the two lists separately, so `Save all` applies an UPDATE where the user staged a DELETE.

`if !self.is_pending_delete(row) { self.row_states.insert(row, RowState::Dirty); }` keeps both the restored override and the pending delete, which is the policy the rest of this file already follows.

crates/dbflux_components/src/components/data_table/model.rs
```diff
@@ -1088,6 +1144,56 @@ mod tests {
         );
     }
 
+    #[test]
+    fn clear_cell_drops_one_staged_value_and_is_undoable() {
```
0xErwin1, line 1152: The buffer test covers the sibling, undo and redo cases well. It is missing the transition the new undo arm gets wrong: stage a cell, `mark_for_delete` on that row, `clear_cell`, then `undo`, and assert `is_pending_delete(0)` still holds.

crates/dbflux_components/src/components/data_table/state.rs
```diff
@@ -962,6 +962,30 @@ impl DataTableState {
...
+        col: usize,
+        cell_value: super::model::CellValue,
+    ) {
+        let base_text = self
+            .model
+            .cell(base_idx, col)
+            .map(|cell| cell.display_text().to_string())
+            .unwrap_or_default();
+
+        if cell_value.display_text().as_ref() == base_text.as_str() {
```
0xErwin1, line 969: Comparing display text keeps the editor's existing semantics and is what fixes the issue, so no change requested here. Worth knowing that it makes the equality textual on every path now: a column shown as `1.50` and a typed `1.5` are different strings, so that pair still stages an edit that a value comparison would treat as a revert. Pre-existing for the editor, now shared by enum, Set NULL, Set default and paste.

0xErwin1 (commented):
The seam is the right call, and it is complete: grepping `crates/` for `set_cell(`, `set_insert_cell(` and `clear_cell(` shows the only `set_cell` left outside the definition is the `else` branch of `stage_base_cell_value` at `state.rs:985`, so no staging path can bypass it now. I did not build or run the branch; this review is static against `e1fb670e` and the diff.

One change I would ask for before merge: the `undo` arm for `ClearCell` writes `RowState::Dirty` unconditionally at `model.rs:928`, which clobbers a pending delete. The other two arms that touch row state, `clear_cell` at 545 and `redo` at 1038, both guard with `is_pending_delete`, so I read this as an oversight rather than a decision. The exact reproduction is on that line.

The rest is coverage: the paste fix has no test at all, and the panel level staging paths that now share the seam have none either. The four context menu call sites also repeat the same visual to source resolution, which is the shape that produced the paste bug in the first place.

`CHANGELOG.md` untouched is right, and `ui:bug` plus `kind:sql` are the right labels.

crates/dbflux_components/src/components/data_table/state.rs
```diff
@@ -1466,6 +1485,65 @@ mod tests {
         (state, input, window)
     }
 
+    /// Regression: typing back the value the row already holds must drop the
+    /// pending change instead of leaving the row marked as modified.
+    #[gpui::test]
+    fn typing_the_rows_own_value_drops_the_staged_edit(cx: &mut gpui::TestAppContext) {
```
0xErwin1, line 1577: Good end-to-end coverage of the editor path, including the assertion that the value is staged first. `stage_base_cell_value` now has five callers and only this one is exercised here; `apply_enum_selection_at`, `handle_set_null`, `handle_set_default` and `handle_paste` have no test at that level.

crates/dbflux_ui_document/src/data_grid_panel/context_menu/mod.rs
```diff
@@ -2020,14 +2020,32 @@ impl DataGridPanel {
...
+            // The selection carries visual indices while the edit buffer is keyed
+            // by source rows, so the pasted value has to be resolved the way the
+            // other staging paths resolve it.
+            let visual_order = state.edit_buffer().compute_visual_order();
+            let cell_value =
+                dbflux_components::components::data_table::model::CellValue::text(&text);
+
+            match visual_order.get(coord.row).copied() {
+                Some(VisualRowSource::Base(base_idx)) => {
+                    state.stage_base_cell_value(base_idx, coord.col, cell_value);
```
0xErwin1: Resolving through the visual order is right and matches the sibling paths: `selection().active` is visual (`record.rs:405` selects from the visual row renderer) while the buffer is keyed by source rows, so `coord.row` was wrong as soon as a pending insert existed. That is a second bug fixed in this PR, and it has no test, which makes it the only change in the diff whose evidence is reasoning rather than a failing case. If the test harness exposes the clipboard the way `handle_copy` writes it, one test with a pending insert above the selection locks this down; if it does not, say so in the PR description.

Separately, and out of scope here: `handle_copy` joins cells with tabs while paste puts the whole string into a single cell, so the round trip is not symmetric. That deserves its own issue.

crates/dbflux_ui_document/src/data_grid_panel/context_menu/mod.rs
```diff
@@ -2122,10 +2139,12 @@ impl DataGridPanel {
 
             match visual_order.get(row).copied() {
                 Some(VisualRowSource::Base(base_idx)) => {
-                    buffer.set_cell(base_idx, col, cell_value);
+                    state.stage_base_cell_value(base_idx, col, cell_value);
```
0xErwin1: This resolution is now repeated four times: 2038, 2142, 2169 and 2220 all do `compute_visual_order`, match `Base` to `stage_base_cell_value` and `Insert` to `set_insert_cell`, while `apply_enum_selection_at` (`state.rs:892-917`) already resolves visual to source inside `DataTableState`. A `stage_cell_value(visual_row, col, value)` method there would collapse all four and make it impossible for the next path to forget the resolution, which is what happened in paste.

crates/dbflux_components/src/components/data_table/model.rs
xilec: Fixed in `1d17039`, exactly as you wrote it: the undo arm restores the override and leaves `PendingDelete` alone. The full sequence you gave is what the test encodes — thanks for tracing it through `request_save_all`.

crates/dbflux_components/src/components/data_table/model.rs
xilec: `undoing_a_cleared_cell_keeps_the_row_pending_delete` (`1d17039`) runs stage → `mark_for_delete` → `clear_cell` → `undo` and asserts the value is back *and* `is_pending_delete(0)` still holds. It fails on the second assertion without the guard.

crates/dbflux_components/src/components/data_table/state.rs
xilec: No change, as you say. The `1.50` / `1.5` pair is real and now shared by every path; it is the semantics the editor already had, and comparing parsed values instead would be a separate decision about how a cell's value is defined.

crates/dbflux_components/src/components/data_table/state.rs
xilec: Covered in `869cedc`: `staging_paths_write_the_row_the_grid_shows` drives `handle_paste`, `handle_set_default` and `handle_set_null` at the panel level, and `enum_selection_stages_the_row_the_table_shows` covers `apply_enum_selection_at` including the null sentinel. Both run with a pending insert between the base rows, so the visual-to-source translation is exercised rather than assumed.

crates/dbflux_ui_document/src/data_grid_panel/context_menu/mod.rs
xilec: The harness does expose it: the test platform round-trips `App::write_to_clipboard`, so the panel test writes the clipboard, selects the visual row and calls `handle_paste` — no workaround needed. With a pending insert it asserts the value lands on the insert and not on the base row below it, and that assertion fails if the visual index is written straight into the buffer. The `handle_copy` asymmetry is untouched, as you suggest.

crates/dbflux_ui_document/src/data_grid_panel/context_menu/mod.rs
xilec: `DataTableState::stage_cell_value(visual_row, col, value)` owns that resolution now (`58f4fca`). The four call sites use it, and `close_editor` and `apply_enum_selection_at` went through it as well, so there is one implementation and the next path cannot forget it.

xilec:
CI note on `869cedc`: the `Deterministic Tests` job failed in `dbflux_ui_base`, not in this branch —

```
hook_phase_runner::tests::run_hook_phase_continues_in_order_after_warn_and_ignore
test storage runtime: Sqlite { path: "/tmp/dbflux_storage_test_…/dbflux.db",
    source: SqliteFailure(Error { code: DatabaseBusy, extended_code: 5 }, Some("database is locked")) }
```

Nothing in this branch touches `dbflux_ui_base`, the same job passed on `e1fb670`, and locally the test passes 3/3 (`cargo nextest run -p dbflux_ui_base -E "test(hook_phase_runner::tests::run_hook_phase_continues_in_order_after_warn_and_ignore)"`), so I read it as a flake in the test's own storage runtime rather than a regression. Every other check on this PR is green.

I cannot re-run the job from a fork — `gh run rerun` needs admin rights on the repository — so a re-run would be appreciated if you agree it is unrelated.

0xErwin1 (approved):
Approved. The ClearCell undo arm now guards with is_pending_delete, exactly as asked, and undoing_a_cleared_cell_keeps_the_row_pending_delete runs the whole sequence: stage, mark for delete, clear, undo. stage_cell_value owns the visual to source resolution for every staging path, including the four panel call sites and close_editor, and the panel test drives paste through the clipboard with a pending insert sitting between the base rows. The one red check on 869cedca was hook_phase_runner in dbflux_ui_base failing on SQLITE_BUSY from the shared test storage helper, which this branch does not touch; the rerun is green.

Issue #600, opened 2026-09-12
[bug] UI stops redrawing after moving the window between outputs (NVIDIA + niri/Wayland)


xilec:
### DBFlux version

0.7.8 (prebuilt binary via the project's nix flake; `main` pins the same gpui/blade versions; `dbflux --version` prints nothing on this build)

### Operating system

Linux (Wayland)

### OS / distro detail

NixOS 26.05, kernel 6.18; niri 26.04 compositor, native Wayland client (not XWayland); NVIDIA GTX 1650, proprietary driver 595.71.05; 3 monitors (one rotated 90°), scale 1.

### Architecture

amd64 / x86_64

### Affected driver (if any)

N/A — rendering/UI layer, not driver-specific (observed with a PostgreSQL connection open and with none).

### Steps to reproduce

1. Start dbflux as a native Wayland client.
2. Move the window to a different monitor (output change, no size change; observed on niri).
3. Interact with the UI: switch tabs, open dialogs.
4. If nothing is frozen yet, repeat the moves a few times — the bug is intermittent, and the set of affected outputs varies between attempts.
5. Observe the frozen frame; then resize the window slightly.

### Expected behavior

The UI keeps redrawing on every output; swapchain invalidation caused by an output change is recovered from automatically.

### Actual behavior

On some outputs the window stops receiving new frames: input is still processed (state changes happen — tabs are switched logically, sidebar selection moves), but the displayed frame never updates. A manual resize (or floating toggle) immediately repaints and unfreezes the window. The output the window was created on keeps working. Nothing in `RUST_LOG=info,gpui=debug,dbflux=debug`, the compositor journal, or dmesg. Running under XWayland (unsetting `WAYLAND_DISPLAY`) makes the problem go away entirely.

### Logs / stack traces

Clean everywhere — app debug log, compositor journal, and dmesg show no errors. A silent frame-path stall, not a crash.

### Analysis (likely root cause — blade, not dbflux itself)

[...]

PR #639, opened 2026-09-17
refactor(ui): move to gpui-pre and the wgpu renderer

### LINKED ISSUE #601: [feature] Migrate from gpui/gpui-component to gpui-pre and GPUI Kit
### Problem

DBFlux depends on `gpui 0.2.2` and `gpui-component 0.5.0` from crates.io. `gpui 0.2.2` (2025-10-22) is the last official `gpui` release and still renders through `blade-graphics 0.7.1`. Zed has since replaced blade with a `wgpu` renderer (zed-industries/zed#46758), and that change never reached a crates.io `gpui` release.

Staying on the blade renderer has concrete costs:

- #600: on Wayland (niri + NVIDIA), moving the window between outputs leaves the swapchain out of date and the UI stops redrawing until a resize. blade never recreates the swapchain on `VK_ERROR_OUT_OF_DATE_KHR`, and the upstream fix (kvark/blade#266) is unmerged. The wgpu renderer handles `Suboptimal` / `Lost` / `Outdated` by reconfiguring the surface. #600 is getting a short-term patched blade fork; this issue tracks the durable fix.
- We fall behind every gpui and gpui-component improvement made since late 2025.

### Proposed solution

Move to the maintained crates.io snapshots:

- `gpui` -> `gpui-pre` (weekly snapshot of Zed's in-tree gpui, currently `0.3.4`, published by the gpui-component maintainer).
- `gpui-component 0.5.0` -> `gpui-component 0.6.1` (project renamed GPUI Kit, repo `longbridge/gpui-kit`), which already depends on `gpui-pre ^0.3.1`.

A spike in an isolated worktree (dependency swap plus `cargo check --workspace`) gave the following:

**Dependency and bootstrap changes**
- `gpui = { package = "gpui-pre", version = "0.3.4" }` keeps `use gpui::...` paths compiling.
- `Application::new()` is gone. Platform backends live in separate crates (`gpui-pre-linux`, `gpui-pre-wgpu`, `gpui-pre-macos`, `gpui-pre-windows`); the binary must add `gpui-pre-platform` and start with `gpui_platform::application()`.
- New transitive dependency: `wgpu 29`.

**gpui-component 0.6 breaking changes that touch DBFlux**
- `InputState` split into `Input` / `Textarea` / `Editor`; `.code_editor()` and `.multi_line()` removed.
- `Table` renamed to `DataTable`; `Divider` renamed to `Separator`; `is_eof` replaced by `has_more` (default flipped).
- Dock API redesigned (`DockItem` -> `DockLayout`, `TabPanel` -> `TabGroup`, `StackPanel` / `Dock` removed).
- `Dialog::new(window, cx)` -> `Dialog::new(cx)`; theme struct restructured; tree-sitter highlighter API changed (per-language features); `ScrollbarShow` removed.

**Measured impact**
- `dbflux_components` fails with 54 errors (E0599 x17, E0061 x13, E0609 x9, E0308 x6, E0432 x4, E0063 x4, E0027 x1). The six downstream UI crates are blocked behind it and still unmeasured.
- Most frequent patterns: renamed theme fields, `LanguageConfig::new` removed, `FocusHandle::focus` now takes `cx`, `code_editor()` removed, `gpui::Corner` gone, `BoxShadow` requires `inset`, chart engine call signatures changed.

**Suggested plan**
1. Bootstrap and dependency swap behind a branch; get `dbflux_components` compiling.
2. Migrate the downstream UI crates one PR at a time (`dbflux_ui_base`, `dbflux_ui_document`, `dbflux_ui_sidebar`, `dbflux_ui_windows`, `dbflux_ui`).
3. Verify Linux (Wayland and X11), macOS, and Windows builds and runtime, including the static MSVC CRT link and the Nix closure (`vulkan-loader` is already present).
4. Drop the patched blade fork once the migration lands and confirm #600 stays fixed.

### Alternatives considered

- **Stay on `gpui 0.2.2` with a patched blade fork.** Fixes #600 in the short term but leaves us on a renderer Zed has abandoned, with a fork to maintain.
- **`gpui-ce`** (community fork, `gpui-ce 0.2.2`, also wgpu-based with the surface fix). Rejected because `gpui-component` targets `gpui-pre`, not `gpui-ce`; switching would mean dropping or forking gpui-component.
- **Depend on Zed's gpui by git revision.** Official source, but leaves crates.io and does not match what gpui-component 0.6 builds against.

Risk to keep in mind: `gpui-pre` is a single-maintainer republish of Zed internals, not an official Zed release. DBFlux already depends on the same maintainer through gpui-component.

### Primary area

ui:feature

### Labels I plan to apply

- [ ] Kind label (e.g. `driver:feature`, `ui:feature`, `query:feature`, `mcp:feature`, `rpc:feature`, …)

0xErwin1:
## Summary

Moves DBFlux off `gpui 0.2.2` + `gpui-component 0.5.0`, which render through `blade-graphics 0.7.1`, and onto `gpui-pre 0.3.5` + `gpui-component 0.6.1`, which render through wgpu. The patched `blade-graphics` fork that `main` carries today is deleted.

The acceptance bar is behavioural parity, not a redesign. No user-facing flow changes by design; the diff is API adaptation plus one deliberate widget-type change described below.

**Review path.** Read the three commits that carry a decision — `build(deps)`, `feat(editor)`, and `refactor(ui_document): drop the suppress_dirty mechanism`. Everything else is a rename, a signature change, or the `AsyncApp::update` adaptation applied crate by crate. The `Cargo.lock` accounts for about 2100 of the 3600 changed lines.

## What does this resolve?

- Resolves #601

Related: #600 is fixed by the same renderer change, but `#602` already carries the fork-based fix in `main`. Either path delivers #600 in 0.8.0; this one removes the fork instead of maintaining it. Worth deciding separately whether #600's user-visible fix should get its own `## [Unreleased]` entry, since it currently has none.

## How was this solved?

Five dependency edits resolve the tree:

| Change | Why |
|---|---|
| `gpui = { package = "gpui-pre", version = "=0.3.5" }` | keeps every existing `use gpui::…` path compiling |
| `gpui-component = { version = "=0.6.1", features = ["tree-sitter"] }` | the highlighter's `LanguageConfig` is behind that feature |
| `tree-sitter = "0.26"`, `tree-sitter-lua = "0.4.1"` | `tree-sitter` declares `links = "tree-sitter"`, so two versions cannot coexist in one graph |
| `gpui-pre-platform` with `default-features = false, features = ["wayland", "x11"]` | its defaults are empty, and those features are what pull the wgpu renderer on Linux |
| `[patch.crates-io] blade-graphics` removed | the renderer it patched is gone |

`Application::new()` no longer exists, so the binary opens the platform through `gpui_platform::application()`.

Three decisions worth a reviewer's attention:

1. **`AsyncApp::update` no longer returns `Result`**, and this accounts for the majority of the diff — 119 of the 250 errors in `dbflux_ui_document` alone. `AsyncApp` holds `std::rc::Weak<AppCell>` and is therefore `!Send`, while `BackgroundExecutor::spawn` requires `Send`, so an `AsyncApp` can never be captured by a background task and every such call runs on the foreground executor inside a `cx.spawn` future. That is the case gpui-pre documents as covered by its executor liveness check, so the "context was gone" error arms were dropped rather than worked around. `AsyncWindowContext::update` and `WeakEntity::update`/`read_with` still return `Result` and were left alone, including every call site reached through `spawn_in`.
2. **Single-line completion fields became `EditorState`.** In 0.6.1 only `EditorState` can host a `CompletionProvider`: `InputBaseState` and `handle_action_for_context_menu` are crate-private, `InputMode` is not multi-line, and `TextareaState` documents that code-editor facilities do not exist on it. The data-grid filter input and the visual query builder's column, table, join and sort inputs are configured to look and behave like one-line fields — no line numbers, no soft wrap, no folding, no search bar, one 24px row, and `submit_on_enter` so plain Enter emits `PressEnter` without inserting a newline while an open completion menu consumes Enter first. With the engine attaching the arrow-key listeners itself, `completion_input_keys_wrapper` has no callers and is gone.
3. **`suppress_dirty` was removed** because 0.6.1's `set_value` is silent, so programmatic loads no longer dirty the buffer and the flag that compensated for it is dead. Programmatic loads stay clean, user edits still dirty the document.

Tradeoff to weigh: the review is large (~3600 source lines across 145 files) and cannot be split. The dependency swap is atomic — `dbflux_components` exposes GPUI types in its public API, so a workspace with some crates on `gpui` and some on `gpui-pre` cannot typecheck — which means `main` cannot take this in stages, and intermediate commits in this branch do not build. Please apply `size:exception` if that is acceptable, or reject the approach and I will rework it.

## Validation

| Gate | Command | Result |
|---|---|---|
| Workspace | `cargo check --workspace --all-targets` | 0 errors |
| Lint | `cargo clippy --workspace --locked --features "sqlite,postgres,mysql,mssql,mongodb,redis,dynamodb,cloudwatch,influxdb,clickhouse,turso,aws" -- -D warnings` | exit 0 (same command as `style.yml`) |
| Format | `cargo fmt --all --check` | clean |
| Tests | `cargo test --workspace --locked --features "…"` | 3 failed, see below |
| Build + run | `cargo build -p dbflux --features …`, then the binary with an isolated `XDG_DATA_HOME` | launches, renders through wgpu/Vulkan, shuts down cleanly |

Runtime evidence from the migrated binary: `gpui_wgpu::wgpu_context` enumerates five adapters including `NVIDIA GeForce GTX 1650 Ti with Max-Q Design (Vulkan, DiscreteGpu)`; the IPC socket binds; the database bootstraps and applies its 31 migrations; `SIGTERM` runs all five shutdown phases and completes in 200ms; no `ERROR` lines. Visual parity against a `main` build was checked and showed no difference.

**Known limitations, not hidden:**

- **Three tests fail**, all in `dbflux_ui_windows connection_manager::form::tests`, all from gpui-pre's leaked-handle detector at `gpui-pre-0.3.5/src/app/entity_map.rs:1116`. Every behavioural assertion passes; the panic is in teardown, with two strong handles still alive to the same `ConnectionManagerWindow`. They reproduce under both `cargo test` and `cargo nextest`. This must be fixed before merge.
- **macOS, Windows, X11 and the Nix closure are unverified** on this branch.
- **#600 was not re-verified on NVIDIA hardware.** gpui-linux selects the compositor's GPU hint, which on the test machine is the Intel iGPU, so the reporter's configuration may not reproduce locally.

## Where was this tested?

- [x] Automated tests
- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] X11
- [x] Wayland
- [ ] Other:

## Checklist

- [ ] I added or updated tests when needed

## Labels to apply

- Kind: `ui:feature`
- Platform: `platform:linux`
- Size: `size:exception` — the dependency swap is atomic, so this cannot be split into chained PRs


xilec (commented):
Thanks for driving this migration — the three decision commits read well, and I double-checked the mechanical adaptation against the gpui-pre / gpui-component sources. Everything held up except one thing I'd hold the merge on.

## The Ayu palette never reaches gpui-component 0.6.1 widgets

In 0.6.1 widgets paint from the `Theme.tokens` snapshot (`Button::Primary` → `tokens.button_primary`, `checkbox.rs` → `tokens.primary`), which only `apply_config` derives — from the *registry default* palette. `apply_ayu_*` mutates the flat fields after `Theme::change`, and neither `sync_base` nor a token recompute happens anywhere in the PR (upstream documents this in `theme/mod.rs`: call `Theme::sync_base` after mutating).

Reproduced at runtime with the test under the expander, following the app's exact init path: `theme.primary` = `Hsla{h: 0.094, s: 1.0, l: 0.665}` (the Ayu accent lands in the flat palette ✔), but `tokens.button_primary.color` and `tokens.primary.color` = `Hsla{h: 0.0, s: 0.0, l: 0.98}` — the near-white upstream default. So primary buttons, checkbox/switch fills and outline hovers render near-white on Ayu dark. This contradicts the "visual parity" validation in the description — could you re-check with a screenshot of any modal's primary button?

<details>
<summary>Repro test — drop into <code>crates/dbflux_components/tests/theme_token_parity_review.rs</code> and run <code>cargo test -p dbflux_components --test theme_token_parity_review</code></summary>

```rust
//! Review-time check for PR #639 (NOT for merging).
//!
//! gpui-component 0.6.1 widgets paint from the `Theme.tokens` snapshot, which
//! `apply_config` derives only from the registry default palette. DBFlux's
//! `apply_ayu_*` mutates the flat `ThemeColor` fields afterwards. These tests
//! assert the user-visible parity contract: what a `Button::Primary` /
//! `Checkbox` paints must be the Ayu accent, i.e. the token snapshot must
//! carry the same colors DBFlux wrote into the flat palette.

use dbflux_components::theme;
use dbflux_core::{AppStyle, ThemeSetting};
use gpui::TestAppContext;

fn apply_dark(cx: &mut gpui::App) {
    gpui_component::init(cx);
    theme::init_with_settings(ThemeSetting::Dark, AppStyle::Default, cx);
}

#[test]
fn primary_button_token_carries_ayu_accent() {
    let cx = TestAppContext::single();
    cx.update(|cx| {
        apply_dark(cx);

        let theme = gpui_component::theme::Theme::global(cx);
        assert_eq!(
            theme.primary, theme.tokens.button_primary.color,
            "Button::Primary paints cx.theme().tokens.button_primary; after \
             apply_ayu_dark it must equal the Ayu accent written into \
             theme.primary (#FFB454), not the upstream default palette"
        );
    });
}

#[test]
fn primary_token_carries_ayu_accent() {
    let cx = TestAppContext::single();
    cx.update(|cx| {
        apply_dark(cx);

        let theme = gpui_component::theme::Theme::global(cx);
        assert_eq!(
            theme.primary, theme.tokens.primary.color,
            "widgets reading cx.theme().tokens.primary (checkbox/switch fills, \
             outline hovers) must see the Ayu accent, not the stale snapshot"
        );
    });
}

#[test]
fn sync_base_alone_is_not_enough_but_token_recompute_fixes_it() {
    let cx = TestAppContext::single();
    cx.update(|cx| {
        apply_dark(cx);

        // Theme::sync_base rebuilds the Base layer (scrollbars, text-view
        // defaults) but leaves the tokens snapshot field untouched.
        gpui_component::theme::Theme::sync_base(cx);
        let theme = gpui_component::theme::Theme::global(cx);
        assert_ne!(
            theme.primary, theme.tokens.primary.color,
            "documents the pitfall: sync_base alone must not satisfy the parity contract"
        );

        // Minimal recovery: re-derive the snapshot from the mutated palette.
        let theme = gpui_component::theme::Theme::global_mut(cx);
        theme.tokens = gpui_component::theme::ThemeTokens::from(theme.colors.clone());
        drop(theme);

        let theme = gpui_component::theme::Theme::global(cx);
        assert_eq!(
            theme.primary, theme.tokens.primary.color,
            "re-deriving fixes tokens backed by fields apply_ayu_* writes"
        );
        assert_ne!(
            theme.primary, theme.tokens.button_primary.color,
            "but not button_primary: the Ayu palette never writes the flat \
             button_primary field the token is copied from"
        );

        // Full recovery additionally requires writing the semantic field
        // families 0.6.1 widgets consume (button_* here).
        let accent = theme.primary;
        let theme = gpui_component::theme::Theme::global_mut(cx);
        theme.colors.button_primary = accent;
        theme.tokens = gpui_component::theme::ThemeTokens::from(theme.colors.clone());
        drop(theme);
        gpui_component::theme::Theme::sync_base(cx);

        let theme = gpui_component::theme::Theme::global(cx);
        assert_eq!(theme.primary, theme.tokens.button_primary.color);
    });
}
```

</details>

The test also shows the fix is bigger than one call: `sync_base` alone doesn't touch the stale snapshot, and re-deriving `tokens = ThemeTokens::from(colors)` fixes `primary` but not `button_primary` — the Ayu palette never writes the flat `button_*` family the tokens copy from. Either write the token-backed field families in `apply_ayu_*`, re-derive the snapshot and call `sync_base` — or register Ayu as a `ThemeConfig` in the `ThemeRegistry` and let `apply_config` derive everything.

## Non-blocking

- **[P1] `ctrl-space` is a dead binding.** `input.rs:40` binds `TriggerCompletion`, but its only handler went with `completion_input_keys_wrapper` (a1e4d8c), and 0.6.1 has no native ctrl-space binding. Remove the binding or wire a handler on `EditorState`.
- **[P2] Shift+Enter now submits in form-navigation fields.** 0.6.1 added the `shift` flag and a shift-enter binding; the `PressEnter { secondary: false, .. }` patterns (`form_navigation.rs:128`, grid filter) match it, so Shift+Enter now exits edit-mode / applies. If that's intended, please document the new behavior (a line in the changelog is enough); otherwise narrow the patterns to `shift: false`.
- **Nits:** 10 unused `AsyncUpdateResultExt` imports in `dbflux_ui_document` plus the now-userless trait (masked by the crate-level `#![allow(unused_imports)]`, which is why CI stays green); `form.rs:1980/2016/2066` tests discard the still-fallible `Result` from `window.update`; `cancel_and_wait` (`hook_phase_runner.rs:72`) kept its `Result` after its only error source was removed.

Issue #651, opened 2026-09-18
[bug] Unused i18n key save_script_failed


xilec:
Found in the post-merge review of #625.

### DBFlux version

0.8.0-dev.0 (built from source, `main` @ 1df7232)

### Operating system

Other

### OS / distro detail

Not platform-specific.

### Architecture

amd64 / x86_64

### Affected driver (if any)

None — i18n catalogs

### Steps to reproduce

1. `grep -rn save_script_failed crates --include='*.rs'`

### Expected behavior

Every catalog key has a production call-site, or is removed together with its catalog entries.

### Actual behavior

The only reference to `document.code.file_ops.error.save_script_failed` is the key-resolution test `file_ops_keys_resolve_in_both_locales` in `crates/dbflux_ui_document/src/code/file_ops.rs`; no production call-site remains after #625 routed failures through `auto_save_failed` / `save_failed`. The key (and its per-locale catalog entries) is dead weight, and the test even pins it as resolving.

### Logs / stack traces

Not applicable — behaviour described from source.

### Labels I plan to apply

- [x] Kind label (e.g. `driver:bug`, `ui:bug`, `query:bug`, `mcp:bug`, `rpc:bug`, …) — `ui:bug`
- [ ] Driver label if applicable (`driver:postgres`, `driver:mongodb`, …)
- [ ] Platform label if platform-specific (`platform:linux`, `platform:macos`, `platform:windows`)
- [ ] Data-model label if relevant (`kind:sql`, `kind:document`, `kind:kv`, `kind:log`)

Filed from a fork, so the reporter cannot apply labels — asking a maintainer to apply `ui:bug` during triage.

Issue #650, opened 2026-09-18
[bug] Bootstrap test depends on the real user data directory


xilec:
Found in the post-merge review of #625.

### DBFlux version

0.8.0-dev.0 (built from source, `main` @ 1df7232)

### Operating system

Other

### OS / distro detail

Not platform-specific — test-environment dependency.

### Architecture

amd64 / x86_64

### Affected driver (if any)

None — app bootstrap test

### Steps to reproduce

1. Run `test_healthy_bootstrap_records_no_scripts_directory_diagnostic` in an environment where `dirs::data_dir()` is missing, unwritable, or unusual (minimal CI sandbox, unusual `XDG_DATA_HOME`).

### Expected behavior

The test is hermetic: it builds an in-memory `StorageRuntime`, so it should not depend on the host's real data directory.

### Actual behavior

`crates/dbflux_app/src/app_state/mod.rs:3158` asserts `state.scripts_directory().is_some()` ("the test environment resolves a scripts directory") — an assertion against the real `dirs::data_dir()`-backed directory, so the test fails on hosts where that directory cannot be resolved or created, rather than on a regression it was written to catch.

### Logs / stack traces

Not applicable — behaviour described from source.

### Labels I plan to apply

- [ ] Kind label (e.g. `driver:bug`, `ui:bug`, `query:bug`, `mcp:bug`, `rpc:bug`, …) — none of the area labels fits a test-only issue in `dbflux_app`; the generic `bug` is the closest
- [ ] Driver label if applicable (`driver:postgres`, `driver:mongodb`, …)
- [ ] Platform label if platform-specific (`platform:linux`, `platform:macos`, `platform:windows`)
- [ ] Data-model label if relevant (`kind:sql`, `kind:document`, `kind:kv`, `kind:log`)

Filed from a fork, so the reporter cannot apply labels — asking a maintainer to apply the generic `bug` during triage.

Issue #649, opened 2026-09-18
[bug] Scripts-folder diagnostics use hardcoded English strings


xilec:
Found in the post-merge review of #625.

### DBFlux version

0.8.0-dev.0 (built from source, `main` @ 1df7232)

### Operating system

Other

### OS / distro detail

Not platform-specific.

### Architecture

amd64 / x86_64

### Affected driver (if any)

None — app state / diagnostics

### Steps to reproduce

1. Switch the UI locale (e.g. Spanish).
2. Trigger the scripts-directory diagnostic (make the scripts directory unwritable so startup records a diagnostic).

### Expected behavior

The message and its suggested action are localized through `dbflux_i18n`, like the neighbouring file-ops messages.

### Actual behavior

`crates/dbflux_ui_base/src/app_state_entity.rs:58-69` — both the diagnostic message ("Scripts cannot be saved to the scripts folder.") and the `suggested_action` text are hardcoded English strings, so non-English users get an English toast/audit entry.

### Logs / stack traces

Not applicable — behaviour described from source.

### Labels I plan to apply

- [x] Kind label (e.g. `driver:bug`, `ui:bug`, `query:bug`, `mcp:bug`, `rpc:bug`, …) — `ui:bug`
- [ ] Driver label if applicable (`driver:postgres`, `driver:mongodb`, …)
- [ ] Platform label if platform-specific (`platform:linux`, `platform:macos`, `platform:windows`)
- [ ] Data-model label if relevant (`kind:sql`, `kind:document`, `kind:kv`, `kind:log`)

Filed from a fork, so the reporter cannot apply labels — asking a maintainer to apply `ui:bug` during triage.

Issue #648, opened 2026-09-18
[bug] A close during an in-flight save writes the same bytes twice


xilec:
Found in the post-merge review of #625; the maintainer asked for a repro, so this issue ships with a deterministic one.

### DBFlux version

0.8.0-dev.0 (built from source, `main` @ 1df7232; test verified against that tree)

### Operating system

Other

### OS / distro detail

Not platform-specific — shared write-queue logic.

### Architecture

amd64 / x86_64

### Affected driver (if any)

None — code editor / file persistence

### Steps to reproduce

User-level: type in a file-backed script, press Ctrl+S, and close the tab before the write lands. The window is short manually, so the deterministic form is the test below — it is the same interleaving the existing test `an_explicit_save_landing_before_the_queued_close_flush_still_closes_the_tab` already exercises; that test's asserted event pair ("the explicit save and the flush that follows each report once") is this issue visible in-tree.

Deterministic repro (drop into the `tests` module of `crates/dbflux_ui_document/src/code/file_ops.rs`; passes on current `main`, demonstrating the double write):

```rust
/// One Ctrl+S plus one close writes the same bytes twice.
///
/// A close gesture that arrives while an explicit save is still in flight
/// sees the dirty flag (it is cleared only when the write lands), so it
/// queues a `CloseFlush` carrying the very bytes the running save is
/// already writing. That flush then re-runs the full staged replacement —
/// create, chmod, rename — for identical bytes. Each `SaveFinished(true)`
/// is emitted only by a completed physical write, so two of them mean two
/// writes of the same content.
#[gpui::test]
fn a_close_during_an_in_flight_explicit_save_writes_identical_bytes_twice(
    cx: &mut TestAppContext,
) {
    let path = temp_save_as_path("redundant-close-flush");

    with_file_backed_document(
        cx,
        path.clone(),
        "OLD;",
        |doc, _app_state, events, window| {
            window.update(|window, cx| {
                doc.update(cx, |document, cx| {
                    document.editor.input_state.update(cx, |state, cx| {
                        state.set_value("NEWEST;", window, cx);
                    });
                    // Ctrl+S; on the test executor the write cannot land
                    // before the synchronous close below runs.
                    document.save_file(window, cx);
                    assert_eq!(
                        document.resolve_close(window, cx),
                        CloseDisposition::Deferred,
                        "the close defers behind the running save"
                    );
                });
            });
            window.run_until_parked();

            assert_eq!(
[...]

PR #652, opened 2026-09-18
fix(editor): ask before closing an untitled buffer and report refused saves

### LINKED ISSUE #640: [bug] A file-backed tab with an untrustworthy baseline can never be closed
Found in the post-merge review of #625. The maintainer is already working on a fix (see the PR thread) — this issue exists so the fix has tracking to close against.

### DBFlux version

0.8.0-dev.0 (built from source, `main` @ 1df7232)

### Operating system

Other

### OS / distro detail

Not platform-specific — shared editor/file-persistence logic.

### Architecture

amd64 / x86_64

### Affected driver (if any)

None — code editor / file persistence

### Steps to reproduce

1. Open a file-backed script and edit it.
2. Quit DBFlux so the session shadow differs from the on-disk file.
3. Modify or delete the file externally.
4. Relaunch: session restore deliberately leaves the baseline unseeded when the physical read fails or the recovered shadow differs from disk (`crates/dbflux_ui/src/ui/views/workspace/actions/query.rs:486`).
5. Close the tab — the write refuses as `BaselineUnknown` and the tab stays open behind an "Auto-save failed" toast.
6. Repeat step 5 — every close gesture repeats the same cycle.

### Expected behavior

A way out of the loop: for example a `suggested_action` on the refusal ("use Save As to keep these edits"), or any other documented escape hatch.

### Actual behavior

`resolve_close` always queues a conflict-checked `CloseFlush` for file-backed documents (`crates/dbflux_ui_document/src/code/file_ops.rs:350-361`). With no baseline — or a baseline belonging to another path — the write refuses as `BaselineUnknown`, the tab stays open, and the next close gesture re-enters the same cycle, each time with an "Auto-save failed" toast. There is no gesture that closes the tab.

### Logs / stack traces

Not applicable — behaviour described from source; reachability confirmed by the maintainer in the #625 thread.

### Labels I plan to apply

- [x] Kind label (e.g. `driver:bug`, `ui:bug`, `query:bug`, `mcp:bug`, `rpc:bug`, …) — `ui:bug`

Filed from a fork, so the reporter cannot apply labels — asking a maintainer to apply `ui:bug` during triage.

### LINKED ISSUE #641: [bug] Dirty untitled buffers have no "Don't save" option on close
Found in the post-merge review of #625. The maintainer is already working on a fix (see the PR thread) — this issue exists so the fix has tracking to close against.

### DBFlux version

0.8.0-dev.0 (built from source, `main` @ 1df7232)

### Operating system

Other

### OS / distro detail

Not platform-specific — shared editor/close logic.

### Architecture

amd64 / x86_64

### Affected driver (if any)

None — code editor

### Steps to reproduce

1. Open an untitled buffer (a new script/query without a backing file).
2. Type something.
3. Close the tab — a Save As dialog opens.
4. Dismiss the dialog.

### Expected behavior

A way to close the buffer without saving — either the pre-#625 unsaved-changes dialog with a "Don't save" choice, or an explicit discard affordance.

### Actual behavior

The unsaved-changes dialog is suppressed for every code document: `decides_own_close` keys off `has_close_policy()`, and every `CodeDocument` registers `resolve_close`, untitled buffers included (`crates/dbflux_ui/src/ui/views/workspace/actions/documents.rs:597-637`). Closing therefore forces Save As; dismissing the dialog keeps the tab open, and there is no gesture that closes the buffer without saving. The #625 rationale ("their file already holds the edits") does not hold for untitled buffers — there is no file.

If dropping discard is an intentional trade-off, it deserves a line in the docs (`docs/USAGE.md`); if not, untitled buffers should keep the dialog.

### Logs / stack traces

Not applicable — behaviour described from source; confirmed by the maintainer in the #625 thread.

### Labels I plan to apply

- [x] Kind label (e.g. `driver:bug`, `ui:bug`, `query:bug`, `mcp:bug`, `rpc:bug`, …) — `ui:bug`

Filed from a fork, so the reporter cannot apply labels — asking a maintainer to apply `ui:bug` during triage.

### LINKED ISSUE #642: [bug] Close/quit write refusals are reported as "Auto-save failed"
Found in the post-merge review of #625.

### DBFlux version

0.8.0-dev.0 (built from source, `main` @ 1df7232)

### Operating system

Other

### OS / distro detail

Not platform-specific — shared editor/file-persistence logic.

### Architecture

amd64 / x86_64

### Affected driver (if any)

None — code editor / file persistence

### Steps to reproduce

1. Trigger any refused `CloseFlush` or `ShutdownFlush` — for example an external conflict on the file (or an untrustworthy baseline), then close the tab or quit the app.

### Expected behavior

The message describes what actually failed: a close-driven or shutdown-driven flush, not an autosave.

### Actual behavior

All refusal outcomes (`ExternalConflict`, `ExternallyDeleted`, `BaselineUnknown`, and `Failed` for the flush kinds) route through the `auto_save_failed` toast key regardless of `WriteKind` (`crates/dbflux_ui_document/src/code/file_ops.rs:158-232`). Closing a tab or quitting the app therefore surfaces "Auto-save failed" — a message that describes a mechanism the user did not trigger, which also mis-locates the problem during triage.

### Logs / stack traces

Not applicable — behaviour described from source.

### Labels I plan to apply

- [x] Kind label (e.g. `driver:bug`, `ui:bug`, `query:bug`, `mcp:bug`, `rpc:bug`, …) — `ui:bug`

Filed from a fork, so the reporter cannot apply labels — asking a maintainer to apply `ui:bug` during triage.

### LINKED ISSUE #643: [bug] Shutdown flush rewrites the session shadow on every 50 ms poll
Found in the post-merge review of #625.

### DBFlux version

0.8.0-dev.0 (built from source, `main` @ 1df7232)

### Operating system

Other

### OS / distro detail

Not platform-specific — shutdown path is shared.

### Architecture

amd64 / x86_64

### Affected driver (if any)

None — code editor / file persistence

### Steps to reproduce

1. Leave a file-backed document dirty and quit the app.
2. Observe the session shadow file during the 2 s flush budget (e.g. with inotify/strace).

### Expected behavior

The shadow written once (or when its content actually changes).

### Actual behavior

`flush_for_shutdown` calls `flush_auto_save` — a synchronous `std::fs::write` of the whole buffer — on every poll while edits are pending (`crates/dbflux_ui_document/src/code/file_ops.rs:743`). The shutdown loop polls with `POLL_INTERVAL = 50 ms` inside `DOCUMENT_FLUSH_TIMEOUT = 2000 ms` (`crates/dbflux/src/main.rs:64-66`), so a document whose flush takes the full budget gets ~40 identical synchronous shadow writes on the path where the process is trying to finish.

### Logs / stack traces

Not applicable — behaviour described from source.

### Labels I plan to apply

- [x] Kind label (e.g. `driver:bug`, `ui:bug`, `query:bug`, `mcp:bug`, `rpc:bug`, …) — `ui:bug`

Filed from a fork, so the reporter cannot apply labels — asking a maintainer to apply `ui:bug` during triage.

0xErwin1:
## Summary

- Closing a code buffer with no file asks first again, instead of forcing Save As.
- A refused close or quit says so, and points at `Ctrl+s` / **Save File As** as the way out.
- The shutdown flush writes the session artifact once per buffer, not once per poll.

## What does this resolve?

- Resolves #640
- Resolves #641
- Resolves #642
- Resolves #643

All four come from the post-merge review notes on #625 that `xilec` filed as #640-#651. This PR takes the two behavioural findings and the two hygiene items with a real cost; the remaining eight issues are untouched and listed under Known limitations.

## How was this solved?

**One close decision point (`close_tab`).** The gate that decides "does this document persist its own pending edits?" lived only in `close_active_tab`, while the tab-bar close button, middle-click, the context menu and the batch routes reached the funnel `close_tab` directly. A predicate-only change would have restored the dialog for `Ctrl+W` and left every other route forcing Save As, so the gate moved into the funnel itself, where every route already arrives. `PaneHandle` gains an optional per-document predicate (`decides_own_close`) and `has_close_policy` now takes `cx`: a code document decides its own close only while it has a file to persist to. A batch close asks once, in a single dialog carrying every document that needs a decision, and `CloseAllTabs` now routes through the same batch helper instead of duplicating its loop.

**"Don't save" removes the tab directly.** `DiscardAll` used to loop back through `close_tab`. With the dialog restored, that would re-enter the gate it came from and re-open the confirmation forever, and the funnel would have flushed the very edits the user chose to drop. It now goes through `close_tab_now`, which is what a document without a close policy already got.

**`resolve_close` fails closed for an untitled buffer.** It used to start Save As and report `Deferred`. Since the workspace asks before any close route gets that far, reaching it means a caller bypassed the gate, and the honest answer is to keep the tab open. This gives `CloseDisposition::KeepOpen` its first production producer. The pending-Save-As guard is untouched and still wins for a document whose Save As is in flight.

**Refusals are attributed and actionable.** A helper maps the write's kind to the toast key, so a refused `CloseFlush` says the tab stayed open and a refused `ShutdownFlush` says the edits stayed in the session, instead of both reporting an autosave the user never triggered. The three conflict-check refusals (external conflict, externally deleted, unknown baseline) also carry `with_suggested_action`, naming `Ctrl+s` / **Save File As**; an explicit save deliberately overwrites, which is what makes that a real escape from the baseline loop in #640. That is the escape hatch the issue asks for.

**The shutdown artifact is written once per buffer.** `flush_for_shutdown` calls `flush_auto_save` on every poll, and that is a synchronous write of the whole buffer; the loop polls every 50 ms against a 2 s budget, so a dirty document was rewritten around 40 times with identical bytes. The content is now built once, recorded on the session state, and rewritten only when it changed. The artifact write is also extracted from `flush_auto_save` so there is one writer and one place that selects the target.

### Tradeoffs and boundaries

- The gate decides dirtiness from `change_summary`, the same predicate the confirmation dialog has always used, so it needs a real diff to list a document. `restore_dirty` sets `is_dirty` without a diff, so a restored buffer with `is_dirty` and identical text would not be asked about; that state is unreachable in production, because `restore_dirty` is called only for a `FileBacked` tab that has a shadow (`actions/query.rs:493`), and a file-backed document decides its own close. I did not change the predicate to close this, because doing so needs a dialog summary for a document with no diff, which is its own decision.
- No manual GUI pass. The close and autosave flows are covered by tests rather than by hand, as in #625.
- Issue #648 ("a close during an in-flight save writes the same bytes twice") is the one finding from the review I could not confirm by reading, and it is not addressed here.
- The `scripts_root()` cleanup tests in `actions/documents.rs` fail in this sandbox because they write into a read-only `$HOME`. They fail identically on `main`, and are untouched.

## Validation

| Command | Result |
|---|---|
| `cargo nextest run --workspace --no-fail-fast` | 6506 passed, 248 skipped |
| `cargo nextest run -p dbflux_ui -p dbflux_ui_document --test-threads=1` | 1528 passed, 3 skipped |
| `cargo fmt --all` | applied, clean |
| `cargo clippy -p dbflux_ui_document -p dbflux_ui -p dbflux_i18n --all-targets` | adds no warning over the base |

Behavioural evidence, observed before each fix and green after (the repository's RED-first convention):

```
FAIL dbflux_ui_document code::file_ops::tests::an_untitled_buffer_reaching_resolve_close_fails_closed
  assertion `left == right` failed: an untitled buffer must not start a Save As from resolve_close
    left: Deferred   right: KeepOpen

FAIL dbflux_ui ui::views::workspace::tab_close_request_tests::a_dirty_untitled_tab_closed_through_the_funnel_asks_first
  the close must wait for the user's decision

FAIL dbflux_ui ui::views::workspace::tab_close_request_tests::a_batch_close_asks_once_about_all_documents_that_need_it
  a batch with documents that must be asked about opens the confirmation

FAIL dbflux_ui_document code::file_ops::tests::refusal_toasts_name_the_close_and_shutdown_contexts
FAIL dbflux_ui_document code::file_ops::tests::a_close_flush_refused_without_a_baseline_reports_a_refusal
FAIL dbflux_ui ui::views::workspace::tab_close_request_tests::only_code_documents_decide_their_own_close
```

Tests added or changed:

- `only_code_documents_decide_their_own_close` — an untitled code document no longer decides its own close; a file-backed one still does.
- `a_dirty_untitled_tab_closed_through_the_funnel_asks_first` — the tab-bar route asks, the tab stays open, and no Save As flow starts.
- `discarding_a_dirty_untitled_tab_removes_it_without_re_asking` — discard removes the tab, does not re-open the confirmation, and never calls the save picker.
- `a_batch_close_asks_once_about_all_documents_that_need_it` — one dialog for every document that needs one; clean tabs still close.
- `an_untitled_buffer_reaching_resolve_close_fails_closed` — `KeepOpen` and no picker invocation.
- `a_repeated_close_on_an_untitled_buffer_keeps_failing_closed` — replaces `a_second_untitled_close_while_save_as_is_pending_starts_no_second_flow`, whose name and assertions encoded the behaviour this PR removes.
- `refusal_toasts_name_the_close_and_shutdown_contexts` — the three contexts produce three distinct messages, each carrying a suggested action. `file_ops_keys_resolve_in_both_locales` now covers `zh_Hans` too.
- `a_close_flush_refused_without_a_baseline_reports_a_refusal` — one reported refusal, and the foreign file is left untouched.

### Verification honesty

- `cargo clippy --workspace -- -D warnings` is **red on `main` at `1df7232f`**, this PR's base: 11 warnings in `dbflux_ui_document`/`dbflux_ui` under the current toolchain (rust-1.95), four of them in `file_ops.rs` test code that predates this branch. I confirmed the same 11 by running clippy in a clean worktree at the base. This PR adds none and fixes none of them; they need their own change.
- `palette_filtering_large_dataset_completes_within_budget` fails in the default parallel mode of a `-p dbflux_ui` run (a 50 ms wall-clock budget) and passes alone and serially. #625 reported the same test; it predates this branch and is not attributed to this change.
- I did not run the website build or its link checker. No relative links changed.

## Where was this tested?

- [x] Automated tests
- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] X11
- [ ] Wayland
- [ ] Other:

## Checklist

## Labels to apply

`ui:bug`

## Known limitations

- #644, #645, #646, #647, #648, #649, #650 and #651 remain open. Most are small and independent: the `let _ = std::fs::remove_file(shadow)` in `mark_clean`, the synchronous read on the close of an empty tab, the `TabManager::close_*` methods and `charts_dashboards.rs:1126` that still bypass the funnel, the exit log that reports a timeout as a completed flush, the unused `save_script_failed` key, the hardcoded English in the scripts-folder diagnostic string, and the bootstrap test that resolves the real data directory.
- #646 is only partly addressed: the batch routes and `CloseAllTabs` now share the funnel, but `TabManager::close_others`/`close_all`/`close_to_left`/`close_to_right`/`close_active` still remove tabs without consulting the close policy. They have no live callers today.
- Two documents saving to the same path remain uncoordinated, and a refused autosave still reports on every debounce while the user keeps typing. Both are carried over from #625's own limitations.
- `close_after_save` is now armed only on the file-backed path. Nothing else reads it, but a future caller that assumed it was always armed for an interrupted close would be wrong.


xilec:
Reviewed this against #640–#643 (filed from my #625 post-merge notes). The four fixes do what the issues ask:

- The gate in `close_tab` restores ask-first on every close route, including batch closes asking once, and the `SaveSelected` arm still gives an untitled buffer Save As when the user picks Save.
- Refusals now name what asked for the write and carry `Ctrl+S` / Save File As as the suggested action — and since an explicit save deliberately overwrites, that is a real exit from the #640 baseline loop.
- The content-dedup on the shutdown artifact matches #643.

I also verified the `restore_dirty` tradeoff from the description: `actions/query.rs:493` only calls it for `FileBacked` tabs with a shadow, so the dirty-without-a-diff state has no policy-less document in production, as claimed.

Three small things:

1. `crates/dbflux_ui_document/src/code/file_ops.rs:335` — the `resolve_close` doc comment still says "an untitled buffer goes through Save As as it always has", which is the behavior this PR removes. The new `None`-arm comment says the opposite; the method doc contradicts it.
2. `crates/dbflux_ui/src/ui/views/workspace/mod.rs:2698` — `open_dirty_untitled_code_tab` has two doc comments glued together; the first two lines look like a leftover of an earlier wording.
3. `refusal_error_key` maps `Explicit | SaveAs` onto `auto_save_failed`, so a failed explicit `Ctrl+S` still reports "Auto-save failed". Pre-existing and outside #642's scope — noting it only because the helper now codifies the mapping, in case a follow-up should give explicit saves their own key.

0xErwin1:
Both doc nits were real. `resolve_close` still described the Save As path this branch removes, which the new `None` arm contradicts, and the fixture did carry two glued doc comments from an earlier wording. Both are fixed in `33493e0c`, pushed.

On the third one, the behaviour is not quite what the mapping reads as. `execute_write` conflict-checks only `Auto | CloseFlush | ShutdownFlush` and returns `None` for `Explicit | SaveAs { .. }` (`file_persistence.rs:383`), so those kinds can never produce a refusal outcome at all. A failed `Ctrl+S` goes through the `Explicit | SaveAs` arm of the `Failed` match, which reports `save_failed` ("Failed to save file: ..."), not the autosave key. So the `refusal_error_key` arm you read is unreachable and only keeps the match exhaustive, which is exactly the kind of mapping that invites this reading, so I made it say so instead of leaving it silent. Giving explicit saves their own message is still a reasonable follow-up, but it is a separate decision from #642.

Thanks for confirming the `restore_dirty` tradeoff against `actions/query.rs:493`. That was the one part of the description I wanted a second pair of eyes on.

xilec:
All three make sense now — thanks. Point 3 was a misread on my part: I took the `refusal_error_key` hunk out of its enclosing `Failed` match, so I missed that `Explicit | SaveAs` reports through `save_failed` and never reaches the conflict-checked arms. The comment you added on the unreachable arm is exactly the right guard against that reading. Both doc fixes confirmed. Nothing further from my side.

PR #656, opened 2026-09-19
fix(ui): keep data-grid column widths across a reload

### LINKED ISSUE #620: [bug] Data grid resets column widths on sort and every data reload
### Problem

Manually resized column widths in the data grid are lost whenever the grid reloads
data. Most visibly: clicking a column header to sort brings every column back to
its heuristic default width.

### Root cause

Column widths live in `DataTableState`
(`crates/dbflux_components/src/components/data_table/state.rs`). They are
initialized from a header-name-length heuristic and updated by the resize drag via
`set_column_width`.

Every data reload goes through `DataGridPanel::rebuild_table`
(`crates/dbflux_ui_document/src/data_grid_panel/mod.rs`), which drops the old
`DataTableState` entity and creates a fresh one — widths are re-derived from the
heuristic and all user adjustments are gone. Paths that hit this:

- Server-side sort: `handle_sort_request` → `pending.requery` → `apply_table_result` → `rebuild_table`
- Client-side sort and sort clear: `apply_local_sort` / `handle_sort_clear` → `pending.rebuild` → `rebuild_table`
- Pagination, manual refresh, auto-refresh, and filter changes all end in `rebuild_table` too

So this is not sort-specific: any data reload resets widths. The codebase already
acknowledges the pattern for `record_mode` ("Every rebuild — refresh, requery,
sort, filter — creates a fresh DataTableState") and re-applies that flag after
each rebuild; column widths have no such carry-over.

### Proposed solution

Stop recreating the state on reload: swap the model on the existing
`DataTableState` instead. `DataTableState::set_model` already exists but is
currently `#[allow(dead_code)]` — it was evidently intended for exactly this
("after a new query loads new rows") and never wired up.

- Rework `apply_table_result` / `apply_collection_result` (and the local-sort
  path) to update the existing state's model rather than building a new entity in
  `rebuild_table`.
- Audit everything `rebuild_table` intentionally resets today — sort state,
  selection, edit buffers, enum options, pk/fk/readonly column indices — and
  decide per item whether it should survive a reload.
- Column widths (and, as a side effect, selection and scroll position) then
  survive sort, pagination, and refresh automatically.

### Dependencies

Implementation should land after the gpui/gpui-component migration (#601): that
migration rewrites the table component layer (`Table` → `DataTable` rename and
related API changes in `dbflux_components`), so doing this first would mean
redoing it on top.

### Alternatives considered

- **Minimal fix: carry widths across `rebuild_table`** — capture
  (column name, width) pairs from the old state and restore them by name after
  the new state is created, the same pattern as `apply_record_mode`. Small and
  safe, fixes widths only; selection and scroll still reset. A reasonable
  stopgap if the full rework slips.
- **Persist widths to storage** (per profile/table/column in SQLite) —
  complementary follow-up that would also make widths survive an app restart;
  not required for this bug.

### Affected area

- `dbflux_components` (`data_table` state/model)
- `dbflux_ui_document` (`data_grid_panel` reload paths)
- All data sources: tables, collections, query results. Not driver- or
  platform-specific.

xilec:
## Summary

A data reload no longer rebuilds the grid state. `DataTableState` is created on the first result and later results swap the model in place through a new `set_model(model, ModelSwap, cx)`, so user-facing grid state survives.

## What does this resolve?

- Resolves #620

## How was this solved?

`DataTableState::set_model` (previously dead code) now carries column widths over by column title, drops what is keyed by the old row indices (open inline editor and its staged value, pending edits/inserts/deletes, undo history, enum options) and either keeps or drops the cursor via `ModelSwap::{KeepCursor, ResetCursor}`.

`DataGridPanel::rebuild_table` keeps its single entry point but only creates the `DataTableState`/`DataTable`/subscription pair on the first result; afterwards it applies the recomputed pk/fk/readonly/insertable/enum metadata to the existing state (with reset semantics, so stale index sets cannot linger) and rebuilds only the derived views (document tree, card list). `initial_sort` fully describes the sort of the incoming rows, so an absent one clears the header indicator; `TableReload::NewColumns` needs no special case for that.

The reload intent is an explicit `TableReload` on the grid: `Preserve` for sort, local sort and refresh, `ResetRows` for pagination, page-size and filter changes, `NewColumns` for a new query result. `run_table_query`/`run_collection_query` take the flag when they issue the request and restore it on the applying side, so a cancelled or failed request cannot affect the next reload. The relational-filter gate runs after the take and hands the intent back before returning, because its result is applied through the pending-rebuild route. Pagination marks are set only in the arms that issue a request.

Cursor semantics on a sort: the cursor keeps its index, so a server-side sort can leave a different row under it, with the inspector rail following it silently. That matches Excel, and it is the intended behaviour rather than a side effect of preserving the state.

Scope: the diff is ~1.2k changed lines, over the ~400 guidance in CONTRIBUTING (Pull Requests, rule 5). Requesting a `size:exception` — the production change is a few hundred lines and the rest is coverage for the reload paths that hid the bug.

## Validation

- `cargo nextest run -p dbflux_components -p dbflux_ui_document` — 1801 passed, 3 skipped, including 19 new tests.
- Manual pass over the reported flow and its neighbours (sort both server and client side, sort clear, pagination, page size, filter, manual refresh, column re-projection, grouped result, record mode, editing/save, inspector, chart/document views) on a debug build under Wayland.

New coverage: width carried by title across a reorder; dropped/added columns; cursor clamp and reset; column scroll kept on paging and reset only for a new projection; editor closed and staged edits dropped on reload; enum options not leaking; sort and record mode preserved; server sort keeps width, cursor and indicator; an unsorted reload drops the stale indicator; pagination keeps width and drops the cursor; the state entity is reused; a new query result remaps widths by name and drops the sort; a request that never reaches a driver does not leak its intent; the reload's own selection event is left to do the inspector re-snapshot when the rail follows the cursor.

Known limitation: the relational-gate hand-over is covered by reading the flag's lifecycle, not by an end-to-end test — that needs a real SQL profile with a populated FK cache. The filter-Enter path and the scroll position are not asserted by tests. The hand-over also assumes the visual-query path actually starts: a `build_visual_select` failure, or a visual query that bails out on an unresolved execution connection, leaves the restored flag in place for the next unrelated reload to consume (cosmetic — one stray cursor reset).

Follow-up (not part of this PR): `cargo clippy --all-targets` reports pre-existing `#[cfg(test)]` lints in files outside this change (`code/*`, `schema_diff/*`, `labels.rs`).

## Where was this tested?

- [x] Automated tests
- [x] Linux
- [x] Wayland
- [ ] macOS
- [ ] Windows
- [ ] X11

## Checklist

- [ ] I applied the appropriate labels (see [CONTRIBUTING.md](../CONTRIBUTING.md#label-guide)) — opened from a fork, so they are requested below for a maintainer to apply during triage

## Labels to apply

`ui:bug`, `query`

This PR comes from a fork, so the author cannot apply labels. Please add `ui:bug` and `query` during triage.


0xErwin1 (changes requested):
Read `3bdd9049`. The approach is right and it goes further than the issue asked: reusing the state entity fixes widths, cursor and scroll in one move, and making the FK/read-only setters unconditional is exactly the detail that would bite anyone doing this piecemeal. The width carry by title with a `VecDeque` per title also handles duplicate column names, which a plain `HashMap<name, f32>` would not. I checked that the width and cursor tests fail against the old `set_model`, so the coverage is real.

Two things I would want in this PR rather than a follow-up:

**1. Paging resets the horizontal scroll as well** (`state.rs`, `reset_scroll_position`, reached through `ModelSwap::ResetCursor`). `TableReload::ResetRows` covers pagination, page size and filter changes, and the column projection is identical across those, so every page change now sends the user back to column 0. The doc comment says "return to the top", which is the vertical part; the two horizontal lines look like collateral. Dropping them (or renaming the helper to say what it resets) keeps the intent and stops the regression.

**2. `set_model` still emits `SelectionChanged`, and that emission is now live on every reload.** While the function was dead code the emission never ran during a rebuild, and a fresh entity never emitted on construction, so nothing was listening. Now, with the cursor preserved, the panel's own handler runs `open_row_inspector` (which re-resolves FK references) and `request_value_panel` on every sort and refresh. Two asks: the doc comment justifies the emission as "so subscribers can validate that the preserved selection is still valid in the new model", but `clamp_selection` does that internally now, so either state the real reason (the rail follows the reloaded rows, and `apply_table_result` has no other place that re-snapshots the inspector) or drop the emission. And note that `set_result` already re-snapshots the inspector explicitly, so on that path the work happens twice.

Smaller notes:

**The in-flight intent has more than one owner.** `grid_table.reload` is written at four sites, taken in `run_table_query` / `run_collection_query`, handed back by the relational gate, restored in two spawned completions, and consumed by `rebuild_table`. The known limitation you documented, the gate hand-back when the visual query never starts, is one instance of a class: any early return between the mark and the request leaves it set. `go_to_next_page` is the cheapest example, since the mark is set before the source match and the `DataSource::QueryResult` arm issues nothing. It is not reachable today because the pager only renders when `is_paginated`, but the shape is what makes the leak possible. Carrying the intent in the request payload (`PendingRequery` already travels with the request) would collapse four sites into one and delete the limitation.

**`_editing_subs` is not cleared on a reload.** `close_editor` drops `editing_cell`, `cell_input` and `enum_dropdown` but leaves the subscriptions, which are only rebuilt at the start of the next edit. They point at the dropped input and dropdown, so they are inert, but clearing them where the editor is dropped removes an assumption someone has to re-derive later.

**Cursor semantics on a sort.** `KeepCursor` keeps the index, so after a server-side sort the cursor stays on the same screen position and lands on a different row, and with the rail following, the inspector switches rows silently. Excel behaves that way, so I would accept it, but for the in-memory sort the panel already holds `original_row_order` and could remap to the same logical row. Either way it is user-visible and worth a sentence in the body.

**Review size.** The diff is 1069 changed lines and CONTRIBUTING §76 asks for a split above roughly 400 unless a maintainer approves a `size:exception`. The body does not declare it. I would approve the exception, since the production change is small and the rest is coverage for the paths that hid the bug, but it should be requested rather than implicit.

What I did not verify: the suites locally. CI's Deterministic Tests, cross-platform clippy and the live driver suite are green on this commit, and I read the diff rather than leaning on that.

xilec:
Both must-fixes are in, plus the smaller notes I could act on without opening a new seam.

**1. Paging no longer resets the column scroll.** `reset_scroll_position` became `scroll_to_first_row` (vertical handles only) and `ResetCursor` uses it; the horizontal reset moved to a separate `scroll_columns_to_start`, which the panel calls only for `TableReload::NewColumns`, where the projection is genuinely new and a pixel offset means nothing. Covered by `set_model_reset_cursor_keeps_the_column_scroll`.

**2. The emission stays; the doc comment now says why.** The swap can clamp the cursor (`clamp_selection`) or drop it (`ResetCursor`), and the panel's rail re-snapshots from that event — that is the real reason, and it is written down now. The `set_result` path no longer does the work twice: the explicit pass stands down when `reload_refreshed_the_rail` sees the event's handler already ran (rail tracking + a live cursor), and still covers what the event cannot — a rail pointed at its own row while the cursor sits elsewhere, and a rail that must be dropped because the result shrank. The value-panel tail is deliberately untouched: its admission predicate is `value_panel_can_follow`, which refuses to clobber an edited-but-unsaved panel, and folding it into the same guard would have changed that silently.

Smaller notes:

- *In-flight intent*: the marks in `go_to_next_page` / `go_to_prev_page` moved into the arms that actually issue a request, so the `DataSource::QueryResult` arm can no longer leave one behind. The rest of the class stands; carrying the intent in `PendingRequery` is the right end state and I kept it out of this PR to hold the delta reviewable.
- *`_editing_subs`*: cleared in `close_editor`, where the input and dropdown they watch are dropped.
- *Cursor semantics on a sort*: accepted as Excel-like — the cursor keeps its position and a different row sits under it — and now stated in the body.
- *Review size*: requested in the body as `size:exception`, with the split rationale.

Local: `cargo nextest run -p dbflux_components -p dbflux_ui_document` — 1801 passed, 3 skipped; `cargo clippy --workspace -- -D warnings` and `cargo fmt --all -- --check` clean.

0xErwin1 (approved):
Both findings are addressed. `scroll_to_first_row` now leaves the column offset alone for a row-set change and `scroll_columns_to_start` covers the new-projection case, with a test on each side of that split, which is the distinction I was missing when I called the whole reset collateral. `close_editor` drops `_editing_subs`, and the pagination marks moved into the arms that issue a request, so the `QueryResult` arm can no longer leave one behind.

Keeping the emission and instead removing the duplicate re-snapshot in `set_result` reads better than dropping it: the rail's reason to re-snapshot is the rows that were just installed, and the comment now says that instead of bounds validation the state already does itself. Thanks for the cursor-semantics paragraph in the body too, that was the one part of the reload contract I could not read off the code.

PR #658, opened 2026-09-19
feat(results): apply a table's staged edits before its tab closes (#629)

### LINKED ISSUE #629: [feature] Persist staged grid edits when a table tab is closed
### Problem

A table tab with staged, not yet applied grid edits has no way to get them into the database from the close flow. The edits are simply lost:

* `Ctrl+W` opens the unsaved-changes dialog, which does list the tab (the grid reports its pending changes through `change_summary`), but pressing **Save** cannot do anything with them: a data document has no save path, so the tab is counted as unsaveable, a warning toast appears, and the tab stays open. There is no working action in that dialog.
* The tab-bar routes (✕, middle-click, the context menu) remove the tab on the spot and drop the edits with no question at all.

So closing a tab is a dead end: either nothing happens, or the work disappears. The only way out today is to apply the pending changes in the grid *before* closing the tab — and nothing in the close flow says so.

With table data the moment of application is what matters, and forgetting to apply is the costly failure: the edit exists only in the grid until it is written, and the user cannot know from the tab that closing it throws the edit away. That is why the edit has to be applied deliberately, at a moment the user sees, and equally why applying it *behind* their back is wrong — a write to a table is not a side effect of tidying up tabs. Asking before closing is the shape that matches both: the user is told there is something to apply, and decides.

Observed on 0.8.0-nightly+18eec1b, niri/Wayland, PostgreSQL.

### Proposed solution

Ask about pending edits when a table tab closes, and carry them through only if the user says so:

* Closing a table tab that has staged edits — by any route (`Ctrl+W`, ✕, middle-click, the context menu, close others/all) — asks what to do with them, in the spirit of DBeaver's *Persist object changes* prompt: apply them and close, discard them and close, or cancel the close.
* The question names what is pending (the counts the grid already shows), so the choice is informed rather than a generic "unsaved changes" warning.
* If the user chooses to apply, the tab closes only once the mutations have landed, and it stays open with its edits if the write fails.
* With nothing staged, closing stays exactly as it is now — no prompt.
* Applying from here goes through the same mutation path the grid's own apply action uses, so transaction mode, approval policy, and error reporting behave identically to applying the edits by hand.

Implementation note: #625 (open) adds the seam this fits — `PaneHandle::has_close_policy` / `resolve_close` with `CloseDisposition::{CloseNow, Deferred, KeepOpen}`, plus one funnel for every tab-bar close gesture. A data document currently reports `CloseNow`, so the natural shape is a close policy for `DataDocument`: `Deferred` while `MutationExecutor` runs the staged edits, `KeepOpen` when they fail, `CloseNow` when the grid is clean.

### Alternatives considered

**Keep the staged edits and restore them when the table is reopened.** Nothing is lost, but nothing is applied either: the user still has to remember to go back and apply the edit, and the restored state has to be reconciled with a refresh, a reconnect, or a transaction that no longer exists. It moves the moment of application further away instead of making it visible, which is the opposite of what is needed here.

### Primary area

UI — data grid + tab close flow (`dbflux_ui`, `dbflux_ui_document`), not driver-specific.

### Labels I plan to apply

- [x] Kind label (e.g. `driver:feature`, `ui:feature`, `query:feature`, `mcp:feature`, `rpc:feature`, …) — `ui:feature`
- [ ] Driver label if applicable (`driver:postgres`, `driver:mongodb`, …) — not applicable, any driver with a mutable grid benefits
- [ ] Platform label if platform-specific (`platform:linux`, `platform:macos`, `platform:windows`) — not applicable, the behaviour is not platform-specific
- [ ] Data-model label if relevant (`kind:sql`, `kind:document`, `kind:kv`, `kind:log`) — not applicable, `kind:sql` and `kind:document` grids hold staged edits the same way

I am not a maintainer of this repository and cannot apply labels from a fork, so the boxes above state what the issue needs rather than what I can set: please add `ui:feature` during triage.

0xErwin1:
## Summary

Closing a table tab that holds staged, unapplied grid edits used to be a dead end. The unsaved-changes
dialog did list the tab, because the grid reports its pending changes, but its action could not do
anything with a data document: a grid has no file to save, so the tab counted as unsaveable and the
dialog answered with a warning toast while leaving it open. The only way out was to apply the edits from
the grid first, and nothing in the close flow said so.

The dialog now offers the decision this asks for — apply the edits and close, drop them and close, or
cancel — through the surface that already funnels every close gesture.

## What does this resolve?

- Resolves #629

Two things the issue's "Problem" section describes have to be read against `main`: the tab-bar routes
(✕, middle-click, the context menu, close others/all) stopped removing a dirty tab without asking when
#625 landed the close funnel 27 minutes after the issue was filed, and *Don't save* and *Cancel* already
worked. What remained was the dead action, and that is what this PR fixes.

## How was this solved?

Three work-unit commits, in dependency order.

**1. One completion path for the grid's mutation reporting** (`232c0d86`). The builder's bulk
UPDATE/DELETE had three near-identical `cx.spawn` arms, one per execution mode, each with its own copy of
the same completion handling. The mode is resolved once now and every outcome goes through one
`finish_mutation`, which carries why the run started and reports whether the edits landed. That path had
no test coverage at all before this commit: the test that existed simulated the success closure's state
effects instead of driving it.

**2. The grid's apply-for-close seam** (`08dfd29e`). `DataGridPanel::apply_for_close` runs the staged
edits through `DataTableState::request_save_all` — the same call the grid's own **Save all** makes — so
the mutation policy, the table delete confirmation and the error reporting are the button's rather than a
second implementation. The close is what waits: the grid asks for the tab once the batch has nothing left
to run.

**3. The dialog carries the verb, and the workspace routes it** (`3bd27792`). Each entry names the action
its pending edits need, and the footer takes the apply wording when every selected entry applies. The verb
is derived from which close action a pane provides rather than stored beside it, so the word the dialog
shows cannot drift from the work the action does.

Key decisions and tradeoffs:

| Decision | Why |
|---|---|
| Reuse the unsaved-changes dialog instead of a dedicated per-table prompt | It already asks once per batch and already carries each entry's pending counts, so the batch routes (close others/all) come for free. A dedicated prompt per table would need one prompt per pending tab. |
| Derive the verb from the pane's close action | The alternative — storing a verb next to the action — can drift, which is the failure this PR fixes. |
| Keep the file and apply wordings separate in a mixed selection | Each entry's own line names its action; inventing a neutral verb for a rare selection would be worse than the save wording the footer has always used. |
| Keep the danger gate where it already is | A table delete already has its own confirmation, which now runs after the dialog. The close route is therefore never a weaker confirmation than the grid's own button. |
| Drop the close intent at every failure | The batch never chains past a failure, so an intent that outlived one would stay armed over a batch that can never finish — and the next unrelated row save would drain that batch and close a tab nobody asked to close. 22 sites call it. |

Known limitations, and where they come from:

- **Quitting still drops staged grid edits.** `on_window_should_close` starts the shutdown sequence
  immediately, and the flush phase is bounded at 2 s by design while a batch apply is unbounded, so an
  apply cannot ride it. Asking before the shutdown sequence is the fix, and it touches the app lifecycle
  rather than the close flow this issue reports; it is the next PR.
- **A failure in the middle of a batch can leave a partial apply.** The batch has no transaction wrapper
  and stops at the first failure, which is pre-existing behaviour; the tab now stays open with the rows
  that did not land, so nothing is lost silently.
- **Dismissing the table delete confirmation parks the batch.** It clears the confirmation but leaves the
  remaining inserts and updates staged — also pre-existing. The close path resumes that parked batch
  rather than re-requesting the deletes, so a second attempt cannot duplicate them.

## Validation

```
cargo nextest run --workspace --no-fail-fast      → 6537 passed, 1 failed, 248 skipped
cargo test --doc --workspace                      → clean
cargo fmt --all -- --check                        → clean
cargo clippy --workspace --all-targets            → no warning in any file this branch touches
cargo check -p dbflux                             → clean (all driver features)
```

The one failure is `dbflux_ui::ui::views::workspace::actions::tests::palette_filtering_large_dataset_completes_within_budget`,
a wall-clock budget test (51–178 ms against a 50 ms budget). It fails the same way on the base commit
`cc0cb3a4` in an untouched checkout, so it is environmental rather than caused by this change; the same
flake is recorded in `odd/tasks/645-648-close-path.md`.

Tests added (16): the completion path per outcome (landed, matched no rows, failed, cancelled, chunked
failure reporting), the staged-edit close seam (clean grid, an apply that never reached the database, an
apply that landed), the dialog's verb per entry and per selection, and the verb per document type reaching
the action.

## Where was this tested?

- [x] Automated tests
- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] X11
- [ ] Wayland
- [ ] Other:

The behaviour is exercised by tests, not by hand: the new gpui tests drive the grid through
`FakeDriver` and the completion path directly. If you want the manual pass (staged edits, then ✕,
then apply and close against a real PostgreSQL) before merging, say so and I will run it.

## Checklist

- [ ] I verified the change against the affected user flow(s)

The first box is unchecked on purpose: see "Where was this tested?". The architecture paragraph is
updated in `ARCHITECTURE.md` and in both translated counterparts, per the translation contract.

## Labels to apply

`ui:feature` — applied. The change is not driver-specific and not platform-specific; `kind:sql` and
`kind:document` grids hold staged edits the same way, which is why the issue itself asked for neither.

The diff is 1109 insertions and 293 deletions across three commits, above the ~400-line guide. It is split as far as the
dependency between the pieces allows: the completion path, the apply seam, and the visible prompt are
separate reviewable commits, and the third contains the user-visible behaviour and its tests.


xilec (commented):
**1. A delete-only apply never reports completion — the tab stays open with the close intent armed**

`execute_table_deletes` and `commit_bulk_deletes_collection` report a successful bulk delete through their own tail (`mutations.rs:1741-1744` and `1933-1936`):

```rust
if panel.pending_batch_remaining.is_some() {
    panel.process_next_batch_op(cx);
} else {
    panel.pending.refresh = true;
}
```

As far as I can tell, that `else` arm is missing `finish_close_after_apply(cx)`. The three places that do report are `queue_refresh_after_mutation_success` (`:150`), the save tail (`:777`), and the pump tail (`:1537`) — and a delete-only batch reaches none of them: `handle_save_all` parks table deletes on `pending_delete_confirm` (or fires the collection bulk delete) without staging `pending_batch_remaining`, so the drain runs straight into this uninstrumented tail.

Reproduced with a test modeled on `a_staged_edit_that_lands_asks_the_close_to_proceed` (staged delete → `apply_for_close` → `confirm_delete` → `run_until_parked`): the buffer comes back clean (`pending_edit_counts == (0, 0, 0)`) but the emitted events are just `[Focused]` — no `MutationFinished`, no `RequestClose`. Two consequences:

- After **Apply** and a confirmed delete, the tab stays open for no visible reason — the user has to close it again.
- `close_after_apply` stays armed, so the next *unrelated* successful row save or insert reaches `finish_close_after_apply` via `queue_refresh_after_mutation_success` and emits `RequestClose` — closing a tab nobody asked to close. If I read the 08dfd29e commit message right, that is exactly the failure it describes as prevented.

The added tests cover an update-only batch (a batch of one) but not a delete-only one; when I ran that variant against this branch, it failed.

**2. `DocumentEvent::MutationFinished` has no consumer, and its doc comment promises one**

`handle.rs:21-26` says the workspace "uses it to give the keyboard back when the close it was waiting on did not happen". I couldn't find that consumer: the tab-manager relay matches specific variants and falls through `_ => {}` (`tab_manager.rs:324-395`), and nothing in `dbflux_ui` subscribes to `MutationFinished`. Compare the save path, where `TabManagerEvent::SaveFinished { succeeded: false }` gives the keyboard back (`workspace/mod.rs:1386-1392`) — a failed apply has no counterpart. Either the relay + workspace handler are missing, or the event and the comment are ahead of the code?

**3. Nit: `RequestClose` doc comment looks stale**

`handle.rs:30-33` still reads "Emitted only for a save the unsaved-changes dialog interrupted … and only by documents that can save" — grids now emit it from `finish_mutation` / `finish_close_after_apply` too.

Everything else checked out for me: the completion-path consolidation is faithful (the per-mode failure labels survive; `rows_affected > 0` as the landing predicate is right), the abandon sites cover the batch failure paths I traced, the dialog verb derivation and the i18n keys (en/es/zh_Hans) are consistent, and `ARCHITECTURE.md` with both translations follows the contract.

0xErwin1:
All three are right. Fixed in `17f55d63`.

**The delete-only apply.** Confirmed, and the armed intent is the worse half. `handle_save_all` parks a table delete on the confirmation without staging `pending_batch_remaining`, so the pump never runs and both bulk-delete tails ended in `pending.refresh = true` with no report. I instrumented both (table and collection) through `finish_close_after_apply`, and `a_delete_only_apply_that_lands_asks_the_close_to_proceed` reproduces it: with the fix reverted it fails emitting `["focused"]`, which matches what you saw, and with the fix it emits `finished:landed` and `request-close`. Your reading of the 08dfd29e message is right, that consequence was the one thing the commit claimed to prevent.

**`MutationFinished`.** The consumer is missing by design, and the doc comment was the defect. The dialog handler returns focus when it starts the work, and an apply has no Save As picker to dismiss, so a failed apply leaves nothing for a workspace handler to react to. I kept the variant because it is what the tests read to tell a landed run from a silent one. If you would rather have the symmetry than the unused variant, say so and I will add the relay and the handler, or drop it and let the tests assert on `RequestClose` alone.

**`RequestClose`.** Stale, fixed, it names both callers now.

Not blocking anything. One thing worth knowing for the next reproduction: this gpui queues `cx.emit` as an effect, so an emission's consequence is only visible after the effects are pumped. My first version of the delete test asserted inside the same update closure and saw nothing at all, which sent me chasing the plumbing before the queue turned out to be the reason.

PR #659, opened 2026-09-19
feat(ui): derive SQL editor keywords from the tree-sitter grammar

### LINKED ISSUE #619: [feature] SQL editor completion: derive statement keywords from the tree-sitter grammar (e.g. EXPLAIN on PostgreSQL)
### Problem

On a PostgreSQL connection, typing `expl` in the SQL editor suggests `EXISTS` but not `EXPLAIN`. The completion provider uses a single static `SQL_KEYWORDS` list shared by every SQL-ish query language (`dbflux_ui_document/src/code/completion.rs`). This looks like the deliberate "keyword completion deferred" scoping from the 0.6.0/0.7.0 notes rather than an oversight.

Parsing is not the issue: the tree-sitter grammar already handles `EXPLAIN [ANALYZE] [VERBOSE]` as a statement prefix, and dbflux classifies `EXPLAIN` as a read-only metadata statement in `query/safety.rs`. Only the completion word list is missing it (along with `WITH`, `UNION`, `CASE`, `RETURNING`, `VACUUM`, …).

### Proposed solution

Derive the statement-keyword list from the registered grammar at runtime. The editor already parses with tree-sitter-sequel, whose keyword rules are uniformly named (`keyword_select`, `keyword_explain`, …). Iterating `tree_sitter::Language` node kinds and collecting the `keyword_*` names yields the full keyword set with zero maintenance, stays in sync when the grammar crate is bumped, and keys off the language rather than any driver. `dbflux_core` already depends on both crates.

Two known trade-offs: multi-word keywords come out as separate tokens and would need a small merge step (e.g. GROUP + BY → `GROUP BY`); and the general SQL grammar is a union of dialects, so a few suggested words may not be valid on every engine — no worse than the current mixed static list.

If this is welcome, I'm happy to prepare a PR.

### Alternatives considered

Driver-declared keyword catalog — an optional field (e.g. `statement_keywords`) on `EditorLanguageProfile` (or a sibling of `SyntaxInfo` in `DriverMetadata`). Each driver declares its own vocabulary, with the current generic list as fallback for drivers that declare none. The grammar-derived route above covers the same need without per-driver declarations. Another shape entirely — happy to follow your preference.

### Primary area

ui

### Labels I plan to apply

- [x] Kind label (`ui:feature`)
- [ ] Driver label if applicable (`driver:postgres`, …)

xilec:
## Summary

The SQL editor's completion list was one hand-kept `SQL_KEYWORDS` constant shared by every SQL-style language, so keywords the bundled tree-sitter grammar already parses — `EXPLAIN`, `VACUUM`, `WITH`, `RETURNING` — were never suggested. The vocabulary now comes from the grammar itself.

## What does this resolve?

- Resolves #619

## How was this solved?

`dbflux_core::sql_statement_keywords()` walks `Language::node_kind_count()`,
keeps named and visible symbols, strips the `keyword_` prefix, uppercases, and
sorts/deduplicates once behind a `OnceLock` — 356 words against
tree-sitter-sequel 0.3.11. The SQL arm of `QueryCompletionProvider` and the
three `SqlItemSink` call sites read that set plus a curated supplement, merged
once in the UI crate. Ranking, prefix filtering, and the whitespace-trigger
suppression are unchanged.

Two decisions worth naming:

- Hidden `keyword_<rule>_token<N>` symbols (24 of them, e.g.
  `keyword_int_token2` for `integer`) are filtered out by
  `Language::node_kind_is_named`. They are the anonymous arms of a `choice` over
  synonyms and their literal text is not reachable through the tree-sitter API,
  so a rule contributes at most its own name.
- The supplement is wider than the two multi-word entries discussed on the
  issue, because a strictly derived list would silently drop 23 words the
  editor completes today: the aggregate function names (the grammar matches
  them as plain identifiers), the spellings hidden inside a `choice` (`ILIKE`,
  `INTEGER`, `INT4`, `SERIAL2/4/8`, `INT1/2/3/8`, `FLOAT4`), and statements the
  bundled grammar does not define at all (`GRANT`, `REVOKE`, `IDENTITY`,
  `FETCH`, `CUBE`, `ROLLUP`, `GROUPING`). They are grouped by reason in
  `SQL_CURATED_KEYWORDS`, each group carrying its own comment. Happy to trim it
  to `GROUP BY` / `ORDER BY` if you prefer the strict reading.
- The list stays a union of dialects (`VACUUM`, `BIN_PACK` are offered on every
  SQL-style language), as agreed on the issue. The driver-declared catalog is
  not introduced; an optional `EditorLanguageProfile` override remains the seam
  if a driver ever needs to subtract.
- Known limitation: because the set is a dialect union and candidates are
  filtered by prefix only, the list is longer than the old constant and some
  words need four characters before it narrows. Filtering or ranking by the
  cursor's context — clause position, expected token, dialect — would change
  completion substantially and is a piece of work of a different size; it is
  not attempted here.

## Validation

- Keyword set checked by dumping the built list: 356 grammar words + 25 curated
  = 381, no `*_TOKEN*` artifacts, sorted, deduplicated. It covers every driver
  on `QueryLanguage::Sql` (postgres, mysql/mariadb, mssql, sqlite, turso,
  clickhouse, redshift), the CloudWatch and OpenSearch SQL modes, InfluxDB v1,
  and DynamoDB PartiQL; mongodb, redis, Flux, CloudWatch Logs Insights QL,
  OpenSearch PPL, Cypher, and the script languages keep their own vocabularies.
- `cargo nextest run -p dbflux_core -p dbflux_ui_document`: 2518 passed. The 7
  remaining failures are the `python3`-backed hook tests, which fail on `main`
  too because the Nix dev shell carries no `python3` (reproduced in a clean
  checkout of `main`).
- `cargo clippy --workspace --locked --features "sqlite,postgres,mysql,mssql,mongodb,redis,dynamodb,cloudwatch,influxdb,clickhouse,turso,aws" -- -D warnings`
- `cargo fmt --all --check`
- New tests: `sql_statement_keywords_include_grammar_only_words` (`EXPLAIN`,
  `VACUUM`, `WITH`, `RETURNING`), `sql_statement_keywords_are_sorted_and_deduplicated`,
  `sql_statement_keywords_are_display_words` (no hidden synonym arm leaks),
  `sql_completion_offers_grammar_derived_keywords` (the SQL arm offers
  `EXPLAIN`, `GROUP BY`, `COUNT`, `ILIKE`, and no `*_TOKEN*` label), and
  `sql_keyword_candidates_merge_grammar_and_curated_words` (pins all 25 curated
  words).
- Verified manually in a live window on Wayland against a PostgreSQL
  connection: the SQL editor's completion popup offers the new keywords. The
  item list itself is pinned by `sql_completion_offers_grammar_derived_keywords`.

## Where was this tested?

- [x] Automated tests
- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] X11
- [x] Wayland
- [ ] Other:

## Checklist

## Labels to apply

`ui:feature`, `kind:sql`

This PR comes from a fork without write access to `0xErwin1/dbflux`, so the labels could not be applied here — please apply them during triage.


0xErwin1 (approved):

0xErwin1:
Thanks, merged as 224cb46. Deriving the vocabulary from the grammar rather than keeping the list by hand is the right call, and naming the synonym and multi-word limits up front made the review straightforward.

Looking forward to the next one.

PR #662, opened 2026-09-20
feat(ui): toggle line comments in the SQL and script editors

### LINKED ISSUE #661: [feature] Toggle comment in the code editors with Ctrl+/
DRAFT — issue into 0xErwin1/dbflux, from `.github/ISSUE_TEMPLATE/feature_request.yml`

TITLE: [feature] Toggle comment in the code editors with Ctrl+/

---

### Problem

Commenting a block of SQL (or Lua, Python, Bash) in the editor is manual work: every touched line needs the
language's prefix typed in front of it, and later removed one by one. While debugging a long query that is a
constant back-and-forth, and the entry point users expect from other editors — `Ctrl+/` — does nothing here.
Script documents have the same problem with different prefixes (`--`, `#`), so a fix has to know the language.

### Proposed solution

A `Toggle Comment` command bound to `Ctrl+/` (`Cmd+/` on macOS) and listed in the command palette:

- With a selection, every line the selection touches is commented; if every non-blank touched line already
  carries the prefix, the same command removes it instead.
- Blank lines are left alone, indentation is preserved, and the comment prefix follows the document's
  language (`--` for SQL, Influx, and Lua; `//` for MongoQuery, Cypher, and Cql; `#` for RedisCommands,
  Python, and Bash).
- The selection stays on the toggled block and the viewport does not move. With no selection the caret walks
  to the next line, the way Zed does it.
- Read-only documents (routine bodies) ignore the command.

A branch with an implementation is ready (`feat/toggle-comment` in the fork `xilec/dbflux`); the pull request
will follow this issue.

### Alternatives considered

- Block comments (`/* … */`, `--[[ … ]]`) instead of line prefixes — desirability is not in question, they
  should be supported too, but they are harder: the delimiters differ per language, they do not nest, and a
  selection that starts or ends inside a statement or an existing comment has to be repaired. Line prefixes
  cover the immediate need with one implementation driven by `QueryLanguage::comment_prefix()`, so block
  comments are deferred to follow-up work instead of being folded into this change.
- Toggle only the line under the caret, with no selection support — the common case is commenting several
  lines at once, and a second chord for the block case is worse than one command that handles both.

### Primary area

ui (code editors: SQL and Lua/Python/Bash script documents)

### Labels I plan to apply

Labels: `ui:feature`. I do not have write access to the upstream repository, so a maintainer needs to apply it.

xilec:
## Summary

Adds a Toggle Comment command to the code editors, bound to `Ctrl+/` (`Cmd+/` on macOS) and registered in the command palette. It works in the SQL editor and in the Lua, Python, and Bash script editors.

## What does this resolve?

- Resolves #661

## How was this solved?

- `crates/dbflux_ui_document/src/code/comment.rs` (new) holds the toggle: it expands the selection to whole lines, decides comment or uncomment from the prefix of every non-blank touched line (blank lines are ignored and left alone), rewrites the block, and returns the range the editor must leave selected.
- The block is applied with `InputState::replace_text_in_range` over the touched lines only. A whole-buffer `replace_all` drops and rebuilds the editor's syntax highlighter, and on gpui-component 0.6.1 the rebuilt highlighter can end up painting from a tree that does not match the text, which leaves the buffer uncolored after a toggle.
- The caret/selection and the scroll offset are restored around the edit, so the viewport does not jump. With no selection the caret walks to the next line, the way Zed does it.
- The toggle marks its own edit so the Change handler's completion menu plumbing (a cursor move plus a zero-length edit that keeps the menu anchored) does not run on it and swallow the restored selection. That mark is cleared by the Change handler that consumes it, not by the toggle: GPUI delivers `emit` after the outermost update finishes, so clearing it in the toggle would be too early.
- The comment prefix comes from `QueryLanguage::comment_prefix()`, so the command needs no per-language code: `--` for SQL, Influx, and Lua, `//` for MongoQuery, Cypher, and Cql, `#` for RedisCommands, Python, and Bash.
- `crates/dbflux_core/src/keymap_types.rs`, `crates/dbflux_ui_base/src/keymap.rs`, the workspace dispatch, and the palette carry the command; i18n strings and `docs/USAGE.md` (plus `es`/`zh_Hans`) document it; `CHANGELOG.md` has the entry under `## [Unreleased]`.
- Known limitation: gpui-base exposes a single active selection, so with multiple cursors the command acts on the active one and collapses the rest. This is documented on `toggle_comment`.
- Follow-up work: block comments (`/* … */`, `--[[ … ]]`) are not part of this change. They are wanted, but they need per-language delimiters, nesting rules, and repair of selections that start or end inside a statement or an existing comment — the resolved issue records them as deferred.

Diff size: 13 files, 975 added lines — 423 of them are tests in `crates/dbflux_ui_document/src/code/mod.rs`, the rest the new `crates/dbflux_ui_document/src/code/comment.rs`, the keymap and palette wiring, the i18n strings, and the docs. The change is one coherent feature, so it is not split further.

## Validation

- `cargo nextest run -p dbflux_ui_document` — 1375 passed, 3 skipped. Twenty-one of them cover the toggle: prefix per language, blank lines and indentation, CRLF, multibyte offsets and a caret inside multibyte text, a selection starting mid-line, read-only documents, the block-only replacement, the caret walking down, the selection surviving an uncomment, and the syntax highlighter not being rebuilt.
- `cargo clippy -p dbflux_ui_document -p dbflux_components --lib -- -D warnings` is clean; `cargo fmt --all` applied.
- Manual: a five-line SQL buffer toggled with the hotkey and through the palette, then ~500 KB files in SQL, Lua, Python, and Bash — the highlighting survives every toggle, the selection survives uncommenting, and the caret walks to the next line on a bare caret.

## Where was this tested?

- [x] Automated tests
- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] X11
- [x] Wayland
- [ ] Other:

## Checklist

## Labels to apply

`ui:feature`, `query`

The change comes from a fork without write access, so the labels need a maintainer.


0xErwin1 (commented):
Two things to check before this lands; the toggle itself and its tests look done.

**The chord may be unreachable on layouts where `/` needs Shift.** `key_chord_from_gpui` (`crates/dbflux_ui_base/src/keymap.rs:18`) copies `keystroke.key` and the modifiers verbatim, and `KeymapStack::resolve` is a HashMap lookup on the derived `Eq` of `KeyChord`, so `("/", primary)` matches nothing else. On es-AR, de-DE or fr-FR a slash takes Ctrl+Shift+7, and what arrives is either `7` or `/` with `shift` set. We already hit this class of failure for the ctrl-shift-digit chords: GPUI normalizes chords at the platform layer before our matcher sees them, which is why those four are registered through `cx.bind_keys` (the note at `crates/dbflux_ui/src/keymap/actions.rs:71`). I could not reproduce it here, the Linux keyboard mapper is not in the `gpui-pre` sources I have, and your tested column is Linux/Wayland, so this needs one run on a non-US layout. If it fires, the fix is a native `KeyBinding` via `cx.bind_keys` rather than another literal chord: GPUI then normalizes the registered chord and the incoming keystroke the same way. Note there is no escape hatch either, the keybindings screen only lists and warns.

**A toggle can leave the SQL completion menu unrefreshed.** `toggling_comment` skips the whole menu plumbing in the Change handler (`crates/dbflux_ui_document/src/code/mod.rs:634`), including the cursor move that closes an open menu. Uncommenting is a deletion, which is the case that block exists for, so a Ctrl+/ with the menu open may leave it stale. Worth one manual check with the menu open before toggling; if it shows, close the menu before the ranged edit and restore the selection after, instead of skipping the block.

Neither is a blocker on its own, but I would not merge before the layout check.

Separately, the comment on the new `dispatch/query.rs` arm says non-code documents decline the command, while the arm returns `Some(true)` unconditionally, same as its neighbours, so that sentence reads wrong.

xilec:
All three addressed.

Layout: checked on a non-US layout — Russian, where `/` is Shift+`\`. `Ctrl+/`
toggles there. The Linux path already normalizes it: `gpui-pre-linux` derives
`Keystroke::key` from the produced character and clears `shift` for
single-character non-letter keys
(`gpui-pre-linux-0.3.5/src/linux/platform.rs:1187`), so a layout where `/` needs
Shift still arrives as `("/", primary)` and the literal chord matches. My tested
column stays Linux/Wayland, so I cannot speak for the macOS mapper; making the
same physical key work on any layout is not something `Keystroke` can express,
so it belongs in a separate request.

Completion menu: fixed the way you suggested — the cursor move that closes an
open menu now happens in the toggle, before the ranged edit, and only the
reopening stays suppressed. Verified manually with the menu open, commenting and
uncommenting: the menu closes and the selection survives. A test fails if that
cursor move consumes the consume-once flag.

dispatch/query.rs: the sentence is gone.

0xErwin1 (approved):
Thank you.

